PR #4, opened 2016-10-01
Add wrapper functions for text waves

ryotako:
I add wrapper functions to test text waves.
One can write text wave literal in test code.

```igorpro
CHECK_EQUAL_WAVES( somefunction(), {1,2,3} ) // ok
CHECK_EQUAL_WAVES( somefunction(), {"a","b","c"} ) // compile error!

CHECK_EQUAL_TEXTS( somefunction(), {"a","b","c"} ) // ok!
```

Please check it!

### DIFF
--- a/procedures/unit-testing-comparators.ipf
+++ b/procedures/unit-testing-comparators.ipf
@@ -1223,6 +1223,82 @@ Function REQUIRE_EQUAL_WAVES(wv1, wv2, [mode, tol])
 	endif
 End
 
+#if (IgorVersion() >=7.00)
+
+Function WARN_EQUAL_TEXTWAVES(wv1, wv2, [mode])
+	Wave/Z/T wv1, wv2
+	variable mode
+
+	if(ParamIsDefault(mode))
+		EQUAL_WAVE_WRAPPER(wv1, wv2, WARN_MODE)
+	else
+		EQUAL_WAVE_WRAPPER(wv1, wv2, WARN_MODE, mode=mode)
+	endif
+End
+
+Function CHECK_EQUAL_TEXTWAVES(wv1, wv2, [mode])
+	Wave/Z/T wv1, wv2
+	variable mode
+
+	if(ParamIsDefault(mode))
+		EQUAL_WAVE_WRAPPER(wv1, wv2, CHECK_MODE)
+	else
+		EQUAL_WAVE_WRAPPER(wv1, wv2, CHECK_MODE, mode=mode)
+	endif
+End
+
+Function REQUIRE_EQUAL_TEXTWAVES(wv1, wv2, [mode])
+	Wave/Z/T wv1, wv2
+	variable mode
+
+	if(ParamIsDefault(mode))
+		EQUAL_WAVE_WRAPPER(wv1, wv2, REQUIRE_MODE)
+	else
+		EQUAL_WAVE_WRAPPER(wv1, wv2, REQUIRE_MODE, mode=mode)
+	endif
+End
+
+#else
+
+Function WARN_EQUAL_TEXTWAVES(wv1, wv2, [mode])
+	Wave/T wv1, wv2
+	variable mode
+
+	if(ParamIsDefault(mode))
+		EQUAL_WAVE_WRAPPER(wv1, wv2, WARN_MODE)
+	else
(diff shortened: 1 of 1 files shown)


t-b:
@ryotako Thanks for the PR.

I do see your point in Igor Pro 7. The line `CHECK_EQUAL_TEXTS(somefunction(), {"a","b","c"} )` does not compile in IP6 though.

What I don't understand:

The following code is legal

```
	Make/T wv = {"a","b","c"}
	CHECK_EQUAL_WAVES(somefunction(),  wv)
```

but as you found out if you define the wave inline

```
	CHECK_EQUAL_WAVES(somefunction(),  {"a","b","c"})
```

it does not compile anymore.

I've opened a thread at [igorexchange](http://www.igorexchange.com/node/7336).

ryotako:
Thank you for your reply!
I use Igor Pro Version 6.34J and `CHECK_EQUAL_TEXTS(somefunction(), {"a","b","c"} )` can compile(J means Japanese version). It probably depends on the minor version of Igor Pro, unfortunately.
 
If everyone cannot use the wrapper functions such as CHECK_EQUAL_TEXTS, they might not be appropriate for the testing framework...
But I often use text waves for operation in Igor Pro and I have to test it. 
(like this: https://github.com/ryotako/igor-writer/blob/master/test/test_writer_string.ipf)
So I would be glad if these wrapper functions were merged.

About the behavior of the compiler, I have an idea.
In usual, a wave type is not tested until it is used. But in the case of inline definition, a wave type can be determined easily and Igor Pro try to do it. 
```
Function doPrint()
  PrintWave( {0} ) // ok
  PrintWave( {} ) // compile error! 
  // I think Igor Pro must determine the type of wave at inline definition.
  // {} can be regarded as both an empty text wave and an empty number wave.
  // So this causes a compile error. 
End

Function PrintWave(w)
  Wave w
  print w
End
```

t-b:
No reply on the igor exchange thread so far.

Could you:
- Add `/Z` to the function parameter wave declarations
- Rename the functions to `*_EQUAL_TEXTWAVES`
- Add some documentation to the new CHECK function ala the existing of EQUAL_WAVE_WRAPPER.

Thanks!

ryotako:
I added `/Z` to the function parameter wave declarations, then the following became a compile error.
```
CHECK_EQUAL_TEXTWAVES(somefunction(),  {"a","b","c"})
``` 

In Igor Pro 6, Wave/T works, but Wave/Z/T do not.
I also tried in Igor Pro 7. In this case, both of Wave/T and Wave/Z/T are legal.
I don't know why... 

One idea is to use #if pragma.
```
#if NumberByKey("IGORFILEVERSION", IgorInfo(3)) < 7
Function WARN_EQUAL_TEXTWAVES(wv1, wv2, [mode])
	Wave/T wv1, wv2
	...
End
#else
Function WARN_EQUAL_TEXTWAVES(wv1, wv2, [mode])
	Wave/Z/T wv1, wv2
	...
End
#endif

```

It is a little bit awkward but safe as far as possible.
I am afraid that such #if pragma causes a problem about the documentation system.
Please tell me your opinion.

t-b:
@ryotako Do you need `WARN_EQUAL_TEXTWAVES` for both IP6 and IP7? If yes, let's do the `#if ...` spaghetti.

You can use `#if (IgorVersion() >= 7.00)` instead of `NumberByKey` I guess.

With doxygen we only have to keep in mind to document the `#if` part.

ryotako:
Although the behavior of IP7 is right rather than IP6, I need this for IP6 for the time being because some of my procedures do not work in IP7 now...

So, I add `#if (IgorVersion() >= 7.00)` pragma (thanks to your advice!), and make a new commit.
The following changes are made.

- Add version check `#if` pragma 
- Rename ..._EQUAL_TEXTS to ..._EQUAL_TEXTWAVES
- Add a simple documentation for new functions

t-b:
@ryotako I've taken the liberty in squashing both commits into one and adapting the documentation.

And I got a reply from WaveMetrics support saying that the current behaviour in IP7 is correct.

If you are okay with my adaptations we can merge.

ryotako:
Thank you for cleaning the commit history and code.
I have no problem with this version.

Issue #33, opened 2019-02-13
*_WAVE passes with NULL_WAVE and existing wave


t-b:
```
Function MyTest()

	Make/FREE wv
	REQUIRE_WAVE(wv, NULL_WAVE)

End
```

```
•runtest("Procedure", allowDebug=1)
  Start of test "Unnamed"
  Entering test suite "Procedure"
  Entering test case "MyTest"
  Leaving test case "MyTest"
  Finished with no errors
  Leaving test suite "Procedure"
  Test finished with no errors
  End of test "Unnamed"
```

@ukos-git 
Am I using that wrong?

ukos-git:
You are using it correctly. That is definitely a bug. The function was mostly built to check for the other types and `NULL_WAVE` is a very specific case that could also be handled by `WaveExists()`.

Unfortunately any input to the following function will Always be true as the statement in the first bracket will collapse to 0 in any case:

```igor
Function CheckType(type)
	Variable type

	return (type & NULL_WAVE) == NULL_WAVE
End
```

```console
•print CheckType(NULL_WAVE)
  1
•print CheckType(100)
  1
•print CheckType(pi)
  1
```

This also means that  any binary combination of `NULL_WAVE` will remove the `NULL_WAVE` in favor of the other flags, work which is probably what we want:

```igor
•print NULL_WAVE | FREE_WAVE | NUMERIC_WAVE
  33
•print FREE_WAVE | NUMERIC_WAVE
  33
```

I will add a PR.

PR #34, opened 2019-02-13
Null Wave comparison

ukos-git:
NULL_WAVE could not be used for comparison. Any input to CHECK_WAVE() and its equivivalents would always return true since binary comparison of `(type & NULL_WAVE) == NULL_WAVE` will always yield 1.

bug introduced in 06b2e062:
fixes https://github.com/byte-physics/igor-unit-testing-framework/issues/33

To ensure correct behaviour the following two test cases were used:

```igor

Function MyTest()
	Make/FREE wv
	CHECK_WAVE(wv, NULL_WAVE)
End

Function MyTest2()
	WAVE/Z wv
	REQUIRE_WAVE(wv, NULL_WAVE)
End
```


procedures/unit-testing-comparators.ipf
```diff
@@ -489,7 +489,12 @@ static Function TEST_WAVE_WRAPPER(wv, majorType, flags, [minorType])
 	if(type2 > 0 && type2 <= 2)
 		type = type | 2^(type2 + 3)
 	endif
-	result = (type & majorType) == majorType
+
+	if(majorType == NULL_WAVE)
+		result = type == majorType
+	else
+		result = (type & majorType) == majorType
+	endif
```
t-b: Thanks. The fix does solve the issue but I'm wondering if we should not also try to fix some more related issues:
- NULL_WAVE can be combined with other bits, but that makes no sense
- NULL_WAVE can be fed into the minorType as well

The issue with bitmasks and 0 (the current value of NULL_WAVE) is known. Can we solve that by using a different non-null value?

ukos-git: `NULL_WAVE` could of course get its own non-null value but wouldn't it be simpler if we delete that comparison completely here and use `REQUIRE(!WaveExists())` instead?

t-b: We currently use `REQUIRE(!WaveExists)` in production but that is IMHO not very nice and quite unidiomatic. It is also too different compared to `REQUIRE_WAVE(wv, NORMAL_WAVE)`. Or am I totally on the wrong track?

ukos-git: So we need to
* change NULL_WAVE to NOT = 0x00
* do this only for maior type. To avoid clashes with upstream, I recommend a high enough number here in case WM adds anything here. Not sure which number, though.
* collapse everything to NULL_WAVE if the NULL_WAVE bit was set.

ukos-git: I moved the type = NULL_WAVE down and added a check for WaveExists() if both WaveType() functions returned a non-zero value.

NULL_WAVE is now 256

t-b: @ukos-git Thanks. will merge.

PR #63, opened 2019-04-02
Support for tests that run background functions

MichaelHuth:
This PR adds support for tests that need to run background activity before resuming.

change of supposed API:

The registration from a test is done by:

	RegisterToBeMonitoredBkgFunctions(string taskList, variable mode, string reentryFunction)

reentryFunction must have the same format as TEST_CASE_PROTO and be named *_REENTRY.
This allows to chain multiple registrations in subsequent reentry functions.

See example11-background.ipf


MichaelHuth:
@t-b As StructPut/Get is limited to numeric fields only, the secondary option was to make all used local primitive variables globals in the packages DF. What do you think?

t-b:
@MichaelHuth How about keeping all local variables in a structure but implementing the read/write routines ourselves? Maybe later IP versions will have something we can use instead of `StructPut`/`StructGet`. This would also make the RunTest function shorter and nicer.

MichaelHuth:
@t-b Ok.

MichaelHuth:
@t-b I had identified another problematic case:
For a run() function like

```
Function run()

RunTest("RegistersUTFMonitor.ipf")
RunTest("AnotherTest.ipf")

End
```

Before Igor could drop to the command line another call to RunTest would take place while a reentry was registered before.
Covering that case would be a lot of effort, as we would have to buffer the RunTest call and Abort instead. After the background test case finishes we would need to resume the buffered RunTest call and - even worse the function execution in run() (and we can not get the callers function state).

So I added a check that allows only reentries from our own background monitor and aborts gracefully otherwise.

t-b:
@MichaelHuth Could you resolve the conflicts? I'll then take a look.

MichaelHuth:
@t-b Conflicts resolved.

t-b:
Documentation:
- Nicely done!
- Missing in the HTML output is the function documentation for `RegisterUTFMonitor`.

For RunTest we have

```
Function definition of RunTest
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

.. doxygenfunction:: RunTest
```

in basics.rst. Maybe add something similiar to advanced.rst?

- I've added a fixup commit for the example 11 and 12. We now not always timeout, use static background function so that both ipfs can be loaded at the same time.

Code:
- Nicely done!
- Please define and use constants for the AND/OR mode of `RegisterUTFMonitor`
- RegisterUTFMonitor should check that the passed `mode` is one of the defined ones.
- RegisterUTFMonitor should check that taskList is not empty. And please also set BCKG_TaskList at the end of the function like the other globals.

procedures/unit-testing-autorun.ipf
```diff
@@ -69,17 +69,23 @@ static Function AfterFileOpenHook(refNum, file, pathName, type, creator, kind)
 			catch
 				err = GetRTError(1)
 				print "The run() function aborted with an RTE and this can not be handled."
+				QuitOnAutoRunFull()
```
t-b: @MichaelHuth Why quit here and not at line 80 before `End`?

t-b: @MichaelHuth `cmd` is now unused.

procedures/unit-testing-basics.ipf
```diff
@@ -1238,8 +1245,420 @@ static Function ExecuteHooks(hookType, hooks, juProps, name, procWin, [param])
 	endswitch
 End
 
+/// @brief Background monitor of the Unit Testing Framework
+Function $"UTFBackgroundMonitor"(s)
```
t-b: @MichaelHuth I'm suprised that this works. I would be in favour of not doing it though.

procedures/unit-testing-basics.ipf
```diff
@@ -1238,8 +1245,420 @@ static Function ExecuteHooks(hookType, hooks, juProps, name, procWin, [param])
...
+	SVAR/Z tList = df:BCKG_TaskList
+	SVAR/Z rFunc = df:BCKG_ReentryFunc
+	NVAR/Z timeout = df:BCKG_EndTime
+	NVAR/Z mode = df:BCKG_Mode
+
+	if(!SVAR_Exists(tList) || !SVAR_Exists(rFunc) || !NVAR_Exists(mode) || !NVAR_Exists(timeout))
+		print "Fatal: UTFBackgroundMonitor can not find monitoring data in package DF, aborting monitoring."
+		return 2
+	endif
+
```
t-b: @MichaelHuth You can do the timeout check here and call `RunTest`. Do we want a `print` for the timeout case as well? I'm not sure.

procedures/unit-testing-basics.ipf
```diff
@@ -1238,8 +1245,420 @@ static Function ExecuteHooks(hookType, hooks, juProps, name, procWin, [param])
...
+		if(mode & BACKGROUNDMONMODE_OR)
+			result = result | stopState
+		else
+			result = result & stopState
+		endif
+	endfor
+	result = (timeout && datetime > timeout) ? 1 : result
+
+	if(result)
+		RunTest("")
```
t-b: @MichaelHuth How about `RunTest("UNUSED_FOR_REENTRY")` just to not confuse a future reader?

procedures/unit-testing-basics.ipf
```diff
@@ -1311,171 +1730,237 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
...
 					if(shouldDoAbort())
 						// abort condition is on hold while in catch/endtry, so all cleanup must happen here
-						ExecuteHooks(TEST_CASE_END_CONST, procHooks, juProps, fullFuncName, procWin, param = keepDataFolder)
+						ExecuteHooks(TEST_CASE_END_CONST, s.procHooks, s.juProps, s.fullFuncName, s.procWin, param = s.keepDataFolder)
 
-						ExecuteHooks(TEST_SUITE_END_CONST, procHooks, juProps, procWin, procWin)
+						ExecuteHooks(TEST_SUITE_END_CONST, s.procHooks, s.juProps, s.procWin, s.procWin)
 
-						ExecuteHooks(TEST_END_CONST, hooks, juProps, name, NO_SOURCE_PROCEDURE, param = allowDebug)
+						ExecuteHooks(TEST_END_CONST, s.hooks, s.juProps, s.name, NO_SOURCE_PROCEDURE, param = s.allowDebug)
```
t-b: @MichaelHuth Do we need to do a `QuitOnAutoRunFull` here?

procedures/unit-testing-basics.ipf
```diff
@@ -1311,171 +1730,237 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
...
 						return global_error_count
 					endif
 				endtry
 
-				ExecuteHooks(TEST_CASE_END_CONST, procHooks, juProps, fullFuncName, procWin, param = keepDataFolder)
+				NVAR/Z bckgRegistered = dfr:BCKG_Registered
+				if(NVAR_Exists(bckgRegistered))
+					if(bckgRegistered)
```
t-b: @MichaelHuth Why not just `if(reentry)` ?

procedures/unit-testing-basics.ipf
```diff
@@ -1311,171 +1730,237 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
...
+						// save state
+						NewDataFolder $PKG_FOLDER_SAVE
+						DFREF dfSave = $PKG_FOLDER_SAVE
+						SaveState(dfSave, s)
+
+						return RUNTEST_RET_BCKG
```
t-b: +1

procedures/unit-testing-autorun.ipf
MichaelHuth: If I move the Quit to the end then it changes functionality. The idea is that by default Igor quits (for AUTORUN_FULL) when RunTest finishes. However in the case that run() throws an RTE RunTest could be never called, though Igor should still quit.

When I move it to the end like:
```
if(err)
  QuitOnAutoRunFull()
endif
```

Would work too, but it splits the error handling code to too separate locations.

procedures/unit-testing-basics.ipf
MichaelHuth: My original idea was to define a StrConstant for the function name as it is in RunTest also checked as the caller function name. However Igor does not take that.

procedures/unit-testing-basics.ipf
MichaelHuth: Ok I can add a StrConstant, though it should have a content that is non valid as function name, like
:UNUSED_FOR_REENTRY:

procedures/unit-testing-basics.ipf
MichaelHuth: Yes, I missed that return.

procedures/unit-testing-basics.ipf
MichaelHuth: The code location is after the call to testcasefunc, thus a new monitoring could have been registered. So I need to reevaluate the global flag.

reentry just buffers locally the state while the 'entry' takes place until line 1899 where ClearReentrytoUTF() is called.

MichaelHuth:
@t-b updated

procedures/unit-testing-autorun.ipf
t-b: @MichaelHuth Good point!

t-b:
@MichaelHuth  I've proofread the RegisterUTFMonitor documentation and while at it, added a fixup commit.

procedures/unit-testing-basics.ipf
t-b: @MichaelHuth Got it.

Can we make that part of the code easier?

Does introducing

```
Function IsBckgRegistered()
    DFREF dfr = GetPackageFolder()
    NVAR/Z bckgRegistered = dfr:BCKG_Registered
    return NVAR_Exists(bckgRegistered) && bckgRegistered == 1
End

Function ResetBckgRegistered()
    DFREF dfr = GetPackageFolder()
    string/G dfr:BCKG_Registered = 0
End
```

help?

procedures/unit-testing-basics.ipf
```diff
@@ -1238,8 +1247,448 @@ static Function ExecuteHooks(hookType, hooks, juProps, name, procWin, [param])
...
+	elseif(mode == BACKGROUNDMONMODE_AND)
+		result = 1
+	else
+		print "Fatal: Unknown mode set for background monitor"
+		return 2
+	endif
+
+	if(timeout && datetime > timeout)
+		print "UTF background monitor has reached the timeout for reentry"
+		RunTest(BACKGROUNDINFOSTR)
```
t-b: @MichaelHuth Should we return here?

procedures/unit-testing-basics.ipf
```diff
@@ -1238,8 +1247,448 @@ static Function ExecuteHooks(hookType, hooks, juProps, name, procWin, [param])
...
+
+	len = strlen(tasklist)
+	if(!len || numtype(len) == 2)
+		print "Tasklist is empty."
+		Abort
+	endif
+
+	if(!(mode == BACKGROUNDMONMODE_OR || mode == BACKGROUNDMONMODE_AND))
+		print "Unknown mode set"
+	endif
```
t-b: @MichaelHuth Missing abort

procedures/unit-testing-basics.ipf
t-b: ResetBckgRegistered could then also be called by ClearReentrytoUTF().

t-b:
@MichaelHuth I took a stab at converting the hardware tests of MIES using this branch here.

First results look very promising, see branch feature/port-hardware-tests-to-new-utf-bkg-feature.

What is, I think, missing is a visual indication that the reentry function is called.

Something like

```diff
$ git diff .
diff --git a/procedures/unit-testing-basics.ipf b/procedures/unit-testing-basics.ipf
index f264cb3..e0ce492 100644
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -1933,7 +1933,7 @@                           // get Description and Directive of current Function for TAP
                                        // set reentry function
                                        SVAR reentryFuncName = dfr:BCKG_ReentryFunc
                                        FUNCREF TEST_CASE_PROTO TestCaseFunc = $reentryFuncName
-
+                                       printf "Entering reentry \"%s\"\r", reentryFuncName
                                endif
                                try
                                        ClearRTError()
```

maybe?

t-b:
@MichaelHuth 
The following example

```
#pragma rtGlobals=3
#pragma TextEncoding="UTF-8"
#pragma ModuleName=Example11

#include "unit-testing"

static Function ReEntryTask(s)
	STRUCT WMBackgroundStruct &s

	return !mod(trunc(datetime), 5)
End

Function BackgroundTest()

	CtrlNamedBackGround testtask, proc=Example11#ReEntryTask, period=1, start
	RegisterUTFMonitor("testtask", 1, "FirstReentry_reentry")
End

Function FirstReentry_REENTRY()

	Abort
End
```

run with `runtest("example11-background.ipf", allowdebug=1)` gives a null string error in `message`. Not sure if that is related to this PR though.

t-b:
@MichaelHuth Am I doing something wrong when using it with IMs?

```
#pragma rtGlobals=3
#pragma TextEncoding="UTF-8"
#pragma IndependentModuleName=ExampleXX

#include "unit-testing"

static Function ReEntryTask(s)
	STRUCT WMBackgroundStruct &s

	return !mod(trunc(datetime), 5)
End

Function BackgroundTest()

	CtrlNamedBackGround testtask, proc=ExampleXX#ReEntryTask, period=1, start
	RegisterUTFMonitor("testtask", 1, "ExampleXX#FirstReentry_reentry")
End

Function FirstReentry_REENTRY()

	Abort
End
```

gives

```
•runtest("example11-background.ipf", allowdebug=1)
  Start of test "Unnamed"
  Entering test suite "example11-background.ipf"
  Entering test case "BackgroundTest"
  called from BackgroundTest
  Specified reentry procedure has wrong format. The format must be function_REENTRY().
  Encountered "Abort" in test case "BackgroundTest", procedure file "example11-background.ipf"
  Leaving test case "BackgroundTest"
  Failed with 1 errors
  Leaving test suite "example11-background.ipf"
  Test finished with 1 errors
  End of test "Unnamed"
```

t-b:
BACKGROUNDMONMODE_AND/BACKGROUNDMONMODE_OR should not be static. And available in the documentation.

t-b:
```
#pragma rtGlobals=3
#pragma TextEncoding="UTF-8"
#pragma ModuleName=ExampleXX

#include "unit-testing"

static Function TEST_CASE_BEGIN_OVERRIDE(name)
	string name

End

static Function ReEntryTask(s)
	STRUCT WMBackgroundStruct &s

	return !mod(trunc(datetime), 5)
End

static Function BackgroundTesta()

	CtrlNamedBackGround testtask, proc=ExampleXX#ReEntryTask, period=1, start
	RegisterUTFMonitor("testtask", 1, "ExampleXX#FirstReentry_reentry")
	Abort
End

static Function BackgroundTestb()

	CtrlNamedBackGround testtask, proc=ExampleXX#ReEntryTask, period=1, start
	RegisterUTFMonitor("testtask", 1, "ExampleXX#FirstReentry_reentry")
	REQUIRE(0)
End

static Function FirstReentry_REENTRY()

	PASS()
End
```

gives

```
•runtest("example11-background.ipf", allowdebug=1)
  Start of test "Unnamed"
  Entering test suite "example11-background.ipf"
  Entering test case "ExampleXX#BackgroundTesta"
  Encountered "Abort" in test case "ExampleXX#BackgroundTesta", procedure file "example11-background.ipf"
  Entering reentry "ExampleXX#FirstReentry_reentry"
  Leaving test case "ExampleXX#BackgroundTesta"
  Entering test case "ExampleXX#BackgroundTestb"
  0: is false
  Assertion "REQUIRE(0)" failed in line 28, procedure "example11-background.ipf"
  Leaving test case "ExampleXX#BackgroundTestb"
  Failed with 2 errors
  Leaving test suite "example11-background.ipf"
  Test finished with 2 errors
  End of test "Unnamed"
  No saved test state found, aborting. (Did you RegisterUTFMonitor in an End Hook?)
```

which I think is misleading.

Please also add these odd examples to internal_dev.

Thanks!

procedures/unit-testing-basics.ipf
MichaelHuth: Yes

procedures/unit-testing-basics.ipf
MichaelHuth: done

MichaelHuth:
@t-b update

t-b:
@MichaelHuth Thanks. Looks good! I'll leave it open for now as I want a passing MIES run first before merging.

MichaelHuth:
I think there would also be the option to allow users to use framework calls from within the background function. For REQUIRE(x) we could search the StackTrace for the Background function that was a caller and that is in running state, stop it and issue an Abort.
RunTest could test against a set ShouldDoAbort on Reentry.

t-b:
@MichaelHuth 

Thanks for the issue.

The MIES tests did not run through.

I got

```
Start of test "Unnamed"
  Entering test suite "UTF_MultiPatchSeqDaScale.ipf"
  Entering test case "MultiPatchSeqDAScale#MSQ_DS1"
  Available number of specified ITC devices = 1
  ITC Device opened, returned deviceID is 0.
  TEST OVERRIDE ACTIVE
  TEST OVERRIDE ACTIVE
  TEST OVERRIDE ACTIVE
  TEST OVERRIDE ACTIVE
  TEST OVERRIDE ACTIVE
  TEST OVERRIDE ACTIVE
  TEST OVERRIDE ACTIVE
  TEST OVERRIDE ACTIVE
  TEST OVERRIDE ACTIVE
  TEST OVERRIDE ACTIVE
  Fatal: Status of monitored background task Testpulse indicates a function error (FUNCERR).
  TEST OVERRIDE ACTIVE
  TEST OVERRIDE ACTIVE
  TEST OVERRIDE ACTIVE
  TEST OVERRIDE ACTIVE
  TEST OVERRIDE ACTIVE
```

- [ ] Igor was also still running. Should we do a  `QuitOnAutoRunFull()` in the failure cases in the UTF background monitor?

- [ ] I'm registering all potential background tasks and some of them are not started for some test cases.

The code is basically

```
Function ReEntryTask1(s)
	STRUCT WMBackgroundStruct &s

	return !mod(trunc(datetime), 5)
End

Function ReEntryTask2(s)
	STRUCT WMBackgroundStruct &s

	return !mod(trunc(datetime), 5)
End

Function BackgroundTesta()

	CtrlNamedBackGround testtask1, proc=ReEntryTask1, period=1, start
	// ReEntryTask2 not started
	RegisterUTFMonitor("testtask1;testtask2", 1, "FirstReentry_reentry")
End

Function FirstReentry_REENTRY()

	PASS()
End
```

Can we support that? I think we need to drop the error checking for that.

Somehow FUNCERR does not work here

```
CtrlNamedBackGround testtask2, proc=ReEntryTask2, period=1
ctrlNamedBackground testtask2, status
print s_info
  NAME:testtask2;PROC:ReEntryTask2;RUN:0;PERIOD:1;NEXT:0;QUIT:0;FUNCERR:1;
```

MichaelHuth:
Ok for the `return 2` within our Monitor we should also add QuitOnAutoRunFull().


```
Function UserTask(s)
	STRUCT WMBackgroundStruct &s

	return !mod(trunc(datetime), 5)
End

function show()
	CtrlNamedBackGround testtask2, proc=Usertask, period=1
	ctrlNamedBackground testtask2, status
	print s_info
	CtrlNamedBackGround testtask3, proc=UnknownFunc, period=1
	ctrlNamedBackground testtask3, status
	print s_info
end
```

gives

```
NAME:testtask2;PROC:Usertask;RUN:0;PERIOD:1;NEXT:0;QUIT:0;FUNCERR:1;
NAME:testtask3;PROC:UnknownFunc;RUN:0;PERIOD:1;NEXT:0;QUIT:0;FUNCERR:1;
```

And the problem is that from this status I by FUNCERR I can not distinguish between a waiting task and an error task.

However I think the correct solution would be to try to resolve the PROC: procedure ourselves in RegisterUTFMonitor to find out the difference between 'parked' and 'will never run'.

MichaelHuth:
@t-b I pushed a version that tries to resolve the procedure on function error to distinguish between an actual error or a parked task.

t-b:
I've fixed two minor issues, pushed and rebased
```diff
commit 5d91a766eef4ed3d7aae3e2d8418a7b0bc552713 (HEAD -> mh_background)
Author: Thomas Braun <thomas.braun@byte-physics.de>
Date:   Wed Apr 10 22:23:50 2019 +0200

    fixup! Added support for test code running in background functions

diff --git a/procedures/unit-testing-basics.ipf b/procedures/unit-testing-basics.ipf
index 37978f6..c1e381b 100644
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -1826,6 +1826,8 @@   // new var/str must be added to strRunTest and added in SaveState/RestoreState

                if(ParamIsDefault(name))
                        s.name = "Unnamed"
+               else
+                       s.name = name
                endif

                if(ParamIsDefault(testCase))
@@ -1983,16 +1985,13 @@                                         // set reentry function
                                        endif
                                endtry

-                               NVAR/Z bckgRegistered = dfr:BCKG_Registered
-                               if(NVAR_Exists(bckgRegistered))
-                                       if(bckgRegistered)
-                                               // save state
-                                               NewDataFolder $PKG_FOLDER_SAVE
-                                               DFREF dfSave = $PKG_FOLDER_SAVE
-                                               SaveState(dfSave, s)
+                               if(IsBckgRegistered())
+                                       // save state
+                                       NewDataFolder $PKG_FOLDER_SAVE
+                                       DFREF dfSave = $PKG_FOLDER_SAVE
+                                       SaveState(dfSave, s)

-                                               return RUNTEST_RET_BCKG
-                                       endif
+                                       return RUNTEST_RET_BCKG
                                endif

                                ExecuteHooks(TEST_CASE_END_CONST, s.procHooks, s.juProps, s.fullFuncName, s.procWin, param = s.keepDataFolde
r)
```

t-b:
And I also added a fixup which removes the error state checking.

The reason is that in MIES we use background functions with a different signature. See the help of CtrlNamedBackground:

```
You may also specify a user function that takes a user-defined STRUCT as long as the first elements of the structure match the WMBackgroundStruct or, preferably, if the first element is an instance of WMBackgroundStruct. Use the started field to determine when to initialize the additional fields. Your structure may not include any String, WAVE, NVAR, DFREF or other fields that reference memory that is not part of the structure itself.
```

and for that the funcref `func` is not assigned and treated as an error. But it is a valid background function. In principle we could search all procedure files for it and check if it takes a structure only argument. But let's leave it without error checking for now okay?

t-b:
We are nearly there. Only one test case in MIES was still failing.

PR #66, opened 2019-04-07
Update .gitlab-ci.yml

t-b:


ukos-git:
the docu section should get an additional section:

https://github.com/byte-physics/igor-unit-testing-framework/blob/7fcc8d26d057790454509d2e9466e66b69809a7f/.gitlab-ci.yml#L11-L21

```yml
  only:
    changes:
      - docu/sphinx/source/*.rst
      - procedures/*.ipf
```

t-b:
@ukos-git Thx. Will do.

ukos-git:
Is this working as expected? I think it's the question if this is "OR" or "AND" for the tags.

https://github.com/byte-physics/igor-unit-testing-framework/blob/7fcc8d26d057790454509d2e9466e66b69809a7f/.gitlab-ci.yml#L47-L48

t-b:
@ukos-git On second thought. Is that really a gain?

t-b:
The tags are and. But we need a new one so that the shell jobs don't end up in docker. I've added shell as tags.

ukos-git:
> But we need a new one so that the shell jobs don't end up in docker. I've added shell as tags.

sounds more like "or" to me.

t-b:
@ukos-git Ready for review.

.gitlab-ci.yml
```diff
@@ -20,11 +20,51 @@ documentation:
...
+    paths:
+      - tests/JU_*.xml
+      - screenshot/*.png
+
+testing-v8:
+  stage: build
+  tags:
+    - macosx, docker
+  image:
+    name: ${CI_REGISTRY}/internal/docker-igorpro:v8
```
ukos-git: Currently `stable` is the same as `v8` (Igor Pro 8.02) but `stable` will automatically increase to the latest official version available.

ukos-git: We could increase v8 to reflect the same status as `stable` in the registry. That would probably be the way to go.

t-b: Once we have IP9 I want to have an additional job for IP9.

PR #70, opened 2019-04-11
Documentation: Don't delete all running container

t-b:
This is not wanted on a CI box where multiple docker container are
running.

@ukos-git Maybe this is always getting in the way of the unit test runs?

### DIFF
--- a/docu/Makefile
+++ b/docu/Makefile
@@ -1,4 +1,4 @@
-default: docu-sphinx clean
+default: docu-sphinx
 
 build-docker-doxygen: debian
 	@make -C docker/doxygen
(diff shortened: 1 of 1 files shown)


ukos-git:
This is the wrong make that is called here. 

https://github.com/byte-physics/igor-unit-testing-framework/blob/master/docu/sphinx/Makefile

You have to cd one more level into docu/shinx.

t-b:
@ukos-git Good point.

t-b:
@ukos-git Nope. We need to execute doxygen as well.

t-b:
@ukos-git If you push the branch directly to this repo CI will pick it up.

Issue #85, opened 2019-04-23
Better error message for threadsafe test case like looking functions


t-b:
Straight from the MIES tests:

```
#include "unit-testing"

threadsafe Function myTest()

End
```

gives

```
•runtest("procedure")
  Error: Prototype function TEST_CASE_PROTO_DGEN was called.
```

and the user has no clue what is going wrong.

t-b:
@MichaelHuth I think the statement in https://github.com/byte-physics/igor-unit-testing-framework/pull/68/files#r275903221 is not correct as this is exactly the point there this bugs out.

PR #86, opened 2019-04-23
The signature check for tc functions in getTestCaseList checks now checks all tc types

MichaelHuth:
Close https://github.com/byte-physics/igor-unit-testing-framework/issues/85

Previously only multi data test case function signatures were checked.

The problem in 85 was that the list contained a 'classic' test case that did not met the prototype function signature. With the multi data test cases, the code at the end of getTestCasesMatch requires now that the test cases returned by getTestCaseList have valid function signatures.

Before Multi Data TC support no TC function signatures were checked beforehand.


t-b:
@MichaelHuth Thanks.

PR #108, opened 2019-10-29
TestCaseEnd: Silently ignore non-killable working folder

t-b:
The following test case

Function MyTest()
  Make data
  Display data
End

and executing it with default flags results in an error in the
TestCaseEnd hook as the workFolder can not be killed, as the displayed
wave still lives in that.

Just ignore the error as we can not do anything else.


t-b:
@ukos-git @MichaelHuth

PR #110, opened 2019-11-14
WIP: Sort Test Cases By Line Number

### LINKED ISSUE #109: Use a good default execution order of test cases
It seems that the order of test case execution is currently not easy to understand. I expect by default to execute the test cases from the top of the file to the bottom of the file and the test suites to run in the given order.

ukos-git:
After combining the two lists for mulit data and normal test cases, sort
the tests by line number.

Closes #109

### DIFF
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -1292,7 +1292,22 @@ static Function/S getTestCaseList(procWin)
 		testCaseList = testCaseList + testCaseMDList
 	endif
 
+	testCaseList = sortTestCaseList(testCaseList, procWin)
+
 	return CheckFunctionSignaturesTC(testCaseList, procWin)
+End
+
+/// Sort a list of function by line number
+static Function/S sortTestCaseList(testCaseList, procWin)
+	string testCaseList, procWin
+
+	WAVE/T testCases = ListToTextWave(testCaseList, ";")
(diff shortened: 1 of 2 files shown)


t-b:
@ukos-git Thanks. Nice patch.

Is the order of the test suites already correct? I've stumbled upon https://github.com/AllenInstitute/MIES/blob/403c8ec284fa87095c27ea212cf91afc33d68e0f/Packages/Testing-MIES/UTF_HardwareMain.ipf#L41 but now I think I just forgot that `AddListItem` by default adds to the front?

t-b:
@ukos-git I think you need something like

```
#if (IgorVersion() >= 7.0)
    // ListToTextWave is available
#else
/// @brief Convert a string list to a text wave
Function/WAVE ListToTextWave(list, sep)
    string list, sep

    Make/T/FREE/N=(ItemsInList(list, sep)) result = StringFromList(p, list, sep)

    return result
End
#endif
```
for IP6 compatibility. I've dug that out from the mies repo, d8d00fa1 (MIES: Raise minimum Igor Pro version to 7.0, 2016-09-22).

ukos-git:
thanks. The code is not tested, yet.

I am unsure if modules/test suites are handled correctly

There is an `addlistitem` in `CheckFunctionSignaturesTC` btw. which could reverse  sorting again.

MichaelHuth:
Feature was implemented with PR https://github.com/byte-physics/igor-unit-testing-framework/pull/121

Issue #92, opened 2019-04-29
Revise handling of test cases with no assertions


t-b:
Currently we just complain to the history if a testcase did not issue any assertions. But in MIES we just found out, after transitioning to multi data test cases, that some old functions where wrongly treated as test cases.

This could have been avoided if a test case without an assertion would trigger a real error instead of a warning.

PR #123, opened 2020-05-12
Ma issue92 error if no assertion

MathiasAugstein:
In a first step printFailInfo was updated to use the function UTF_PrintStatusMessage (not necessary, but it should make the output more unified) instead of a simple print.

In the second step the error handling in TestCaseEnd when the assertion counter equals 0 was changed. It does not simply print to the console without anything else, but also uses the standard ReportResults to increase the error counter and send messages via PrintFailInfo.

To get the right error message an additional return-case needed to be inserted in getInfo to get the previous message. Furthermore the case "Assertion failed in unknown location" now only triggers if the assertion counter is unequal 0. This should be no problem, because ReportResults is only called from unit-testing-assertion-wrappers.ipf, which only get called from assertions.

The documentation could not be compiled correctly, instead of doxygen input a frame appears "Warning: doxygenfunction: Cannot find function “RunTest” in doxygen xml output for project “igor-utf” from directory: xml". Please check if this is a local problem or unit testing framework documentation is broken.

Closes: #92

### DIFF
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -285,7 +285,7 @@ static Function DebugOutput(str, booleanValue)
 	string &str
 	variable booleanValue
 
-	sprintf str, "%s: is %s", str, SelectString(booleanValue, "false", "true")
+	sprintf str, "%s: is %s.", str, SelectString(booleanValue, "false", "true")
 	if(EnabledDebug())
 		UTF_PrintStatusMessage(str)
 	endif
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -502,7 +502,9 @@ End
 // 0 failed, 1 succeeded
 static Function/S getInfo(result)
 	variable result
-
+	
+	DFREF dfr = GetPackageFolder()
+	NVAR/SDFR=dfr assert_count
 	string caller, procedure, callStack, contents
 	string text, cleanText, line, callerTestCase
 	variable numCallers, i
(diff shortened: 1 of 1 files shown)


MathiasAugstein:
> The documentation could not be compiled correctly, instead of doxygen input a frame appears "Warning: doxygenfunction: Cannot find function “RunTest” in doxygen xml output for project “igor-utf” from directory: xml". Please check if this is a local problem or unit testing framework documentation is broken.

Found the problem, somehow the directory docu/doxygen/doxygen-filter-ipf/ was empty, git pull did not download the content. Anyway, was only a local problem.

t-b:
@MathiasAugstein doxygen-filter-ipf is a git submodule. And these are only correctly checked out when using `git clone --recurse-submodules`. If that has been forgotten you can use `git submodule init` followed by `git submodule update --checkout`.

MathiasAugstein:
> message string should be descriptive

Changed message to e.g.
`Existance of assertion in test case "ErrorWithoutAssertions#EmptyFunction": is false  The test case did not make any assertions!`
Is this too long?

MathiasAugstein:
Now an example error message is
`Test case "ErrorWithoutAssertions#EmptyFunction" contained at least one assertion: is false.  The test case did not make any assertions!`
I globally added the point after "false" or "true", to make it easier to read. Furthermore I decided for "assertions" as last word because of this explanation: https://english.stackexchange.com/a/113005

procedures/unit-testing-basics.ipf
```diff
@@ -502,7 +502,10 @@ End
 // 0 failed, 1 succeeded
 static Function/S getInfo(result)
 	variable result
-
+	
+	DFREF dfr = GetPackageFolder()
+	NVAR/SDFR=dfr assert_count
+	string msg
```
MichaelHuth, line 508: msg is also not used in getInfo

MathiasAugstein:
I also split the last commit, as the change to DebugOutput is global.

Issue #114, opened 2020-01-29
CHECK_EQUAL_WAVES does not detect difference between free and regular waves.


MichaelHuth:
Consider the following test function:

```
Function test()

	Make/T w1 = {"test"}
	Make/FREE/T w2 = {"test"}
	CHECK_EQUAL_WAVES(w1, w2)

End
```

This test runs through without error, but there should be an error like
`Assuming equality using mode <new mode here> for waves w1 and _free_: is false  Assertion "CHECK_EQUAL_WAVES(w1, w2)" failed in line ...`

t-b:
Conclusion: Mention in documentation that the free/normal property of the wave is not checked. Use `*_WAVE` for that.

PR #125, opened 2020-05-14
Ma issue114 docu check equal waves

MathiasAugstein:
In the first commit an error in Example4 was corrected.
After that the description for CHECK_EQUAL_WAVES was changed to state that the function only checks the content.
I tried to add the reference to CHECK_WAVES as clickable link, but it did not work easily.

Closes: #114

### DIFF
--- a/docu/sphinx/source/examples.rst
+++ b/docu/sphinx/source/examples.rst
@@ -139,7 +139,7 @@ Example4
 
 This test suite shows the use of test assertions for waves.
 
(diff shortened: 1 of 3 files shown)


MathiasAugstein:
Link to CHECK_WAVE was added, furthermore CHECK_EQUAL_TEXTWAVES also needed to be changed.

MathiasAugstein:
Docu now states
`Tests two text waves for equality of content (use WARN_WAVE, CHECK_WAVE or REQUIRE_WAVE for type)`
with links at the three *_WAVE.

PR #127, opened 2020-05-18
MA issue10 debugMode option

### LINKED ISSUE #10: Add option to RunTest to break into debugger on error/rte/etc.

MathiasAugstein:
The debugMode option was added which has 3 bits to set:
0x01: debug on RT error
0x02: debug on NVAR SVAR WAVE reference error
0x04: debug on failed assertion

Closes #10 

Open Questions:
 - At the beginning of RunTest(), debugMode is tested to be within 0 and 7. Does it need to be tested for integer values, e.g. by trunk(debugMode) == debugMode ?
- The constants DEBUG_ENABLE and DEBUG_ALLOW have the same value, but are used in different contexts. Is it OK to have two constants?
 - EnableIgorDebugger() gives a 4 bit variable to SetIgorDebugger(), which only uses the 3 least significant bits. Is this a possible bug in the future, so it should only pass 3 bits?

### DIFF
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -365,11 +365,12 @@ static Function RestoreIgorDebugger(debuggerState)
 	SetIgorDebugger(debuggerState)
 End
 
-/// Create the variable igorDebugState in PKG_FOLDER
-/// and initialize it to zero
-static Function InitIgorDebugState()
+/// Create the variables igor_debug_state and igor_debug_assertion
+/// in PKG_FOLDER and initialize it to zero
+static Function InitIgorDebugVariables()
 	DFREF dfr = GetPackageFolder()
-	variable/G dfr:igor_debug_state = 0
+	Variable/G dfr:igor_debug_state = 0
+	Variable/G dfr:igor_debug_assertion = 0
 End
 
 /// Creates the variable status in PKG_FOLDER
@@ -1021,9 +1022,9 @@ End
 
 /// Internal Setup for Testrun
 /// @param name   name of the test suite group
-static Function TestBegin(name, allowDebug)
+static Function TestBegin(name, debugMode)
 	string name
-	variable allowDebug
+	variable debugMode
 
 	variable reEnableDebugOutput, runCountStored
 	string msg
@@ -1056,10 +1057,13 @@ static Function TestBegin(name, allowDebug)
 		EnableDebugOutput()
 	endif
 
-	if(!allowDebug)
-		initIgorDebugState()
-		NVAR/SDFR=dfr igor_debug_state
+	InitIgorDebugVariables()
+	NVAR/SDFR=dfr igor_debug_state
+	if(!debugMode)
 		igor_debug_state = DisableIgorDebugger()
+	endif
+	if(debugMode & (IUTF_DEBUG_ENABLE | IUTF_DEBUG_ON_ERROR | IUTF_DEBUG_NVAR_SVAR_WAVE | IUTF_DEBUG_FAILED_ASSERTION))
+		igor_debug_state = EnableIgorDebugger(debugMode)
 	endif
 
 	string/G dfr:message = ""
@@ -1074,9 +1078,9 @@ End
 
 /// Internal Cleanup for Testrun
 /// @param name   name of the test suite group
-static Function TestEnd(name, allowDebug)
+static Function TestEnd(name, debugMode)
 	string name
-	variable allowDebug
+	variable debugMode
 
 	string msg
 
@@ -1094,7 +1098,7 @@ static Function TestEnd(name, allowDebug)
 	sprintf msg, "End of test \"%s\"", name
 	UTF_PrintStatusMessage(msg)
 
-	if (!allowDebug)
+	if (debugMode != IUTF_DEBUG_ALLOW)
 		NVAR/SDFR=dfr igor_debug_state
 		RestoreIgorDebugger(igor_debug_state)
 	endif
@@ -2219,6 +2223,7 @@ static Structure strRunTest
 	variable enableTAP
 	variable enableRegExp
 	variable allowDebug
+	variable debugMode
 	variable keepDataFolder
 
 	string procWin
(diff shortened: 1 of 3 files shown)


MichaelHuth:
- It does not compile as DEBUG_DISABLE is not defined.

Regarding the constants:
The constants should be non-static that a IUTF user is able to use the constants when calling RunTest.
It would be best to define them in unit-testing-basics.ipf and to avoid collisions give them a prefix like IUTF_*.

It is ok to have two different constants with the same value for two different optional parameters.

I think it is ok, if it is documented for SetIgorDebugger that it evaluates only the 3 LSB.

procedures/unit-testing-basics.ipf
```diff
@@ -2389,16 +2394,22 @@ End
 ///                         The Igor debugger will be left in its current state when running the
 ///                         tests.
 ///
+/// @param	debugMode      (optional) default disabled, enabled when set to 1-7: @n
```
MichaelHuth, line 2404: default disabled means 0, while for the user only 1 - 7 is allowed. It seems to be a bit strange that the default value is not part of the range that the user can set.
It does not hurt to include 0 for DEBUG_DISABLE as it is logical. There should also be something documented (briefly) for the case one uses allowDebug and debugMode at the same time.

procedures/unit-testing-basics.ipf
```diff
@@ -2454,6 +2466,12 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
 			return NaN
 		endif
 
+		if(s.debugMode > 7 || s.debugMode < 0)
+			printf "debugMode can only be an integer between 0 and 7. The input %g is wrong, set to 0.", s.debugMode
+			s.debugMode=0
+		endif
+		s.debugMode = (s.debugMode << 1) | s.allowDebug
```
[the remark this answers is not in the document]
MathiasAugstein: Question: If the option allowDebug=1, should we ignore debugMode (just keep the previous state), or should debugMode be able to overwrite the debugger state?

procedures/unit-testing-basics.ipf
```diff
@@ -2454,6 +2466,12 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
 			return NaN
 		endif
 
+		if(s.debugMode > 7 || s.debugMode < 0)
+			printf "debugMode can only be an integer between 0 and 7. The input %g is wrong, set to 0.", s.debugMode
+			s.debugMode=0
+		endif
```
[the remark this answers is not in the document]
MathiasAugstein: Added the utility function IsInteger, which checks numtype==0 && if var==trunc(var). Hope this is complete.

procedures/unit-testing-basics.ipf
```diff
@@ -307,19 +314,51 @@ static Function DisableIgorDebugger()
...
+///						0x02: debug on error
+///						0x04: debug on NVAR SVAR WAVE reference error
+static Function SetIgorDebugger(state)
+	variable state
+	
+	variable prevState, enable, debugOnError, nvarSvarWave
+
 	DebuggerOptions
-	debuggerState = V_enable
+	prevState = !!V_enable | (!!V_debugOnError) << 1 | (!!V_NVAR_SVAR_WAVE_Checking) << 2
```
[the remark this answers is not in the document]
MathiasAugstein: Instead of multiplying with hard-coded numbers, all multiplications are with the IUTF_DEBUG - constants. Now we should have the freedom to change the bit pattern of debugMode freely.

MathiasAugstein:
debugMode overwrites allowDebug, as it sets a specific state. This is documented under allowDebugs' information.
Manual also looks good, added an example how to use the IUTF_DEBUG - constants (see RunTest parameter list).

procedures/unit-testing-utils.ipf
```diff
@@ -51,6 +51,12 @@ threadsafe static Function IsEmpty(str)
 	return numtype(len) == 2 || len <= 0
 End
 
+/// @brief Returns one if var is a nonfinite integer, zero otherwise
+threadsafe function IsInteger(var)
```
MichaelHuth, line 55: - should be `static`
- `Function`

procedures/unit-testing-basics.ipf
```diff
@@ -2455,6 +2476,13 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
 			return NaN
 		endif
 
+		var = (IUTF_DEBUG_ENABLE | IUTF_DEBUG_ON_ERROR | IUTF_DEBUG_NVAR_SVAR_WAVE | IUTF_DEBUG_FAILED_ASSERTION)
+		if(s.debugMode > var || s.debugMode < 0 || !IsInteger(s.debugMode))
+			printf "debugMode can only be an integer between 0 and %d. The input %g is wrong, aborting!.\n", var, s.debugMode
```
MichaelHuth, line 2496: A hint to use the IUTF_DEBUG constants fits here as well.

procedures/unit-testing-basics.ipf
```diff
@@ -47,6 +47,14 @@ static Constant WAVETYPE1_DFR = 0x03
 static Constant WAVETYPE1_WREF = 0x04
 /// @}
 
+/// @name Constants for Debugger mode
```
MichaelHuth, line 50: a group anchor like
```
/// @anchor DebugConstants
/// @{
/// @}

```

is missing around the block

procedures/unit-testing-basics.ipf
```diff
@@ -2462,6 +2483,17 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
 			return NaN
 		endif
 
+		var = (IUTF_DEBUG_ENABLE | IUTF_DEBUG_ON_ERROR | IUTF_DEBUG_NVAR_SVAR_WAVE | IUTF_DEBUG_FAILED_ASSERTION)
+		if(s.debugMode > var || s.debugMode < 0 || !UTF_Utils#IsInteger(s.debugMode))
+			printf "debugMode can only be an integer between 0 and %d. The input %g is wrong, aborting!.\n", var, s.debugMode
```
MichaelHuth, line 2503: use \r for new line and prefer printf

MathiasAugstein:
Also "variable" and "string" are written as required by our coding conventions.

PR #129, opened 2020-05-20
AreWavesEqual creates better message for wave data check

### LINKED ISSUE #30: Easier out of the box debugging for failed test assertion `*_EQUAL*WAVES`
I just had a failed test with the following JUNIT output:

```
FAIL: Assuming equality using mode 033 for waves guiStateNumRef and guiStateNumNew: is false   Assertion "CHECK_EQUAL_WAVES(guiStateNumRef, guiStateNumNew, mode = WAVE_DATA | DIMENSION_LABELS)" failed in line 230, procedure "UTF_DAEphys.ipf" 
```

and I had to do the following to debug it:
- Rerun the test with `keepDataFolder=1` so that I get access to the waves
- Duplicate the waves and diff them manually
- call `WaveStats` to find the first non-matching entry

This is quite tedious to do manually.

Outputting this kind of info always would be really helpful.

- Output the first non-matching entry together with wave indizes and dimension labels
- Use tolerance correctly
- Investigate if it is feasable to support all modes or only `WAVE_DATA`

MathiasAugstein:
When the data in two waves differ, AreWavesEqual now gives information
where the first and the last difference occured.
FindDifferentWaveEntries, GetWaveIndexInfo and GetWaveDimension funtions
were added in order to work.

Remark: In FindDifferentWaveEntries the built-in function FindLevel was not used, as it returns NAN if already the first element is above the tolerance, or if no element is above the tolerance. That would require an additional step, even though it would be simple if we can assume the second case is impossible because the waves are not equal.

Closes #30

### DIFF
--- a/internal_dev/EqualWaves_Examples.ipf
+++ b/internal_dev/EqualWaves_Examples.ipf
@@ -30,6 +30,13 @@ Function EqualWavesType1()
 	CHECK_EQUAL_WAVES(wv1, wv2, mode=WAVE_DATA)
 End
 
+Function EqualWavesType1_1()
+
+	Make/FREE/N=0/D wv1
+	Make/FREE/N=1/I wv2
+	CHECK_EQUAL_WAVES(wv1, wv2, mode=WAVE_DATA)
+End
+
 Function EqualWavesType2()
 
 	Make/FREE/N=1/D wv1
@@ -61,6 +68,22 @@ Function EqualWaves3()
 	Make/FREE wTest1, wTest2
 	Make/FREE/WAVE wv1 = {wTest1, wTest2}
 	Make/FREE/WAVE wv2 = {wTest2, wTest1}
+	CHECK_EQUAL_WAVES(wv1, wv2, mode=WAVE_DATA)
+End
+
+Function EqualWaves3_1()
+
+	Make/FREE wTest1, wTest2
+	Make/FREE/WAVE wv1 = {wTest1, wTest2}
+	Make/FREE wv2 = {0, 0}
+	CHECK_EQUAL_WAVES(wv1, wv2, mode=WAVE_DATA)
+End
+
+Function EqualWaves3_2()
+
+	Make/FREE wTest1, wTest2
+	Make/FREE/WAVE wv1 = {wTest1, wTest2}
+	Make/FREE/DF wv2 = {NewFreeDataFolder(), NewFreeDataFolder()}
 	CHECK_EQUAL_WAVES(wv1, wv2, mode=WAVE_DATA)
 End
 
--- a/internal_dev/EqualWaves_Examples.ipf
+++ b/internal_dev/EqualWaves_Examples.ipf
@@ -1,0 +1,184 @@
+#pragma rtGlobals=3
+#pragma TextEncoding = "UTF-8"
+#pragma rtFunctionErrors=1
+
+#include "unit-testing"
+
+// All tests here should fail and give an detailed message
+// from CHECK_EQUAL_WAVES in WAVE_DATA mode
+
+Function EqualWavesTol()
+
+	CHECK_EQUAL_WAVES({1}, {1}, mode=WAVE_DATA, tol=NaN)
+End
+
+Function EqualWavesNull()
+
+	WAVE/Z wv1 = $""
+	CHECK_EQUAL_WAVES(wv1, {1}, mode=WAVE_DATA)
+End
+
+Function EqualWavesSize()
+
+	CHECK_EQUAL_WAVES({1}, {2, 3}, mode=WAVE_DATA)
+End
+
+Function EqualWavesType1()
+
+	Make/FREE/N=0/D wv1
+	Make/FREE/N=0/I wv2
+	CHECK_EQUAL_WAVES(wv1, wv2, mode=WAVE_DATA)
+End
+
+Function EqualWavesType2()
+
+	Make/FREE/N=1/D wv1
+	Make/FREE/N=1/T wv2
+	CHECK_EQUAL_WAVES(wv1, wv2, mode=WAVE_DATA)
+End
+
+Function EqualWavesType3()
+
+	Make/FREE/N=1/D wv1
+	Make/FREE/N=1/D/C wv2
+	CHECK_EQUAL_WAVES(wv1, wv2, mode=WAVE_DATA)
+End
+
+Function EqualWaves1()
+
+	CHECK_EQUAL_WAVES({1}, {2}, mode=WAVE_DATA)
+End
+
+Function EqualWaves2()
+
+	Make/FREE/T wv1 = {"1"}
+	Make/FREE/T wv2 = {"2"}
+	CHECK_EQUAL_WAVES(wv1, wv2, mode=WAVE_DATA)
+End
+
+Function EqualWaves3()
+
+	Make/FREE wTest1, wTest2
+	Make/FREE/WAVE wv1 = {wTest1, wTest2}
+	Make/FREE/WAVE wv2 = {wTest2, wTest1}
+	CHECK_EQUAL_WAVES(wv1, wv2, mode=WAVE_DATA)
+End
+
+Function EqualWaves4()
+
+	Make/FREE/DF wv1 = {NewFreeDataFolder()}
+	Make/FREE/DF wv2 = {NewFreeDataFolder()}
+	CHECK_EQUAL_WAVES(wv1, wv2, mode=WAVE_DATA)
+End
+
+Function EqualWaves5()
+
+	Make/FREE/C/D wv1 = {cmplx(Inf, NaN)}
+	Make/FREE/C/D wv2 = {cmplx(NaN, Inf)}
+	CHECK_EQUAL_WAVES(wv1, wv2, mode=WAVE_DATA)
+End
+
+// expect 8 errors in table from 20 to 27, [1][0][2][0] to [2][2][2][0]
+// this also proves the correct indexing order in checking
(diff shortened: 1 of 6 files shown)


MathiasAugstein:
Very early stage, there is no more time for big improvements. Might be finished later.

MathiasAugstein:
Changes were made, now it should work with all types of waves and handle data differences correctly.

As discussed on phone this was intended to be a first version with the main highlight on correct execution. There is a lot of code duplication in the functions for each wave type (e.g. `FindDifferentWaveEntryNumeric`), it should be possible to transfer the 4 loops inside `FindDifferentWaveEntry` and call the corresponding wave-specific functions inside the loops.

MichaelHuth:
Current output design:
```
  Text wave differences:
Dimensions              |DimLabels |Wave Name  |Text          |
------------------------|----------|-----------|--------------|
    0;    0;    0;    0;|LABEL1;;;;|wa in root:|Der eine Text |
                        |LABEL3;;;;|wb in root:|Der andre Text|
    1;    0;    0;    0;|LABEL2;;;;|wa in root:|mehr Text     |
                        |LABEL4;;;;|wb in root:|viel mehr Text|
```

t-b:
That looks like a good start.

MichaelHuth:
@t-b How should we handle the latest IP9 changes regarding the printout? Is there some outlook that a %p format code will be available?

t-b:
> @t-b How should we handle the latest IP9 changes regarding the printout? Is there some outlook that a %p format code will be available?

I'm investigating.

t-b:
@MichaelHuth Could you add some examples to internal_dev?

And I'm also not seeing a detailed message for the following test case:

```
Function Test()

	CHECK_EQUAL_WAVES({1}, {2})
End
```

MichaelHuth:
Hmm I get as JUnit:

```xml
<?xml version="1.0" encoding="utf-8"?>
<testsuites>
	<testsuite package="Procedure" id="0" name="Procedure" timestamp="2021-03-04T02:00:06" hostname="DESKTOP-H4EJ6SM" tests="1" failures="1" errors="0" skipped="0" disabled="0" time="0.014">
		<properties>
			<property name="User" value="Micha"/>
			<property name="System" value="OS:Windows 10 Pro;
OSVERSION:6.3.18363;
LOCALE:US;
IGORFILEVERSION:9.0.0.3"/>
			<property name="Experiment" value="deleteme"/>
			<property name="UTFversion" value="1.08"/>
			<property name="IgorInfo" value="IGORVERS:9.00;
BUILD:36995;
IGORKIND:pro64;
FREEMEM:140731578343424;
PHYSMEM:68607193088;
USEDPHYSMEM:653590528;
NSCREENS:1;
SCREEN1:DEPTH=32,RECT=0,0,3840,2160"/>
		</properties>
		<testcase name="mytest in Procedure (50)" classname="mytest" time="0.008">
			<failure message="Assuming equality using mode WAVE_DATA for waves _free_ and _free_; detailed: Waves difference:
Wave1:0x084767a370
Wave2:0x084767a710
Dimensions  |Labels|Value   |
------------|------|--------|
[0][0][0][0]|      |1.000000|
            |      |2.000000|

: is false.  Assertion &quot;CHECK_EQUAL_WAVES({1}, {2})&quot; failed in line 3, procedure &quot;Procedure&quot;" type="FAIL"></failure>
		<system-out>Entering test case &quot;mytest&quot;
Assuming equality using mode WAVE_DATA for waves _free_ and _free_; detailed: Waves difference:
Wave1:0x084767a370
Wave2:0x084767a710
Dimensions  |Labels|Value   |
------------|------|--------|
[0][0][0][0]|      |1.000000|
|      |2.000000|

: is false.  Assertion &quot;CHECK_EQUAL_WAVES({1}, {2})&quot; failed in line 3, procedure &quot;Procedure&quot;
Leaving test case &quot;mytest&quot;
</system-out>
		<system-err>Assuming equality using mode WAVE_DATA for waves _free_ and _free_; detailed: Waves difference:
Wave1:0x084767a370
Wave2:0x084767a710
Dimensions  |Labels|Value   |
------------|------|--------|
[0][0][0][0]|      |1.000000|
            |      |2.000000|

: is false.  Assertion &quot;CHECK_EQUAL_WAVES({1}, {2})&quot; failed in line 3, procedure &quot;Procedure&quot;</system-err>
		</testcase>
		<system-out>Entering test case &quot;mytest&quot;
Assuming equality using mode WAVE_DATA for waves _free_ and _free_; detailed: Waves difference:
Wave1:0x084767a370
Wave2:0x084767a710
Dimensions  |Labels|Value   |
------------|------|--------|
[0][0][0][0]|      |1.000000|
|      |2.000000|

: is false.  Assertion &quot;CHECK_EQUAL_WAVES({1}, {2})&quot; failed in line 3, procedure &quot;Procedure&quot;
Leaving test case &quot;mytest&quot;
</system-out>
		<system-err>Assuming equality using mode WAVE_DATA for waves _free_ and _free_; detailed: Waves difference:
Wave1:0x084767a370
Wave2:0x084767a710
Dimensions  |Labels|Value   |
------------|------|--------|
[0][0][0][0]|      |1.000000|
            |      |2.000000|

: is false.  Assertion &quot;CHECK_EQUAL_WAVES({1}, {2})&quot; failed in line 3, procedure &quot;Procedure&quot;</system-err>
	</testsuite>
</testsuites>

```

looks solid to me.

t-b:
Jeep, thanks. That error was on my side.

MichaelHuth:
Rebased and with examples

t-b (changes requested):
Code works as expected!

Some output issues:
- Can the dimension sizes and type table, EqualWavesSizes() EqualWavesType*(), have the same format as for different
  data? Basically I want to have the same layout always.
  Including the wave names before the table?
- Add a space after ":" in "Wave1:..."

Code issues:
- Does not compile on IP6
- Other CI failures
- Please use the WAVEYTPE0_XXX/WAVEYTPE1_XXX constants you introduced over magic numbers
- Prefer i, j, k, l as outlined in the coding guidelines for loops variables
- "TODO only IP7" should be fixed or?
- "convert 32 bit int or less to double without precision loss" I'm not understanding that. I thought we mimick the IP behaviour regarding 64bit integers? I. e. casting them down to double?

Edgy edge cases:

The following example (please add it)

```igorpro
Function EqualWavesTextWithNulls()

	string str1, str2
	
	str1 = "abcd"
	str2 = PadString(str1, strlen(str1) + 1, 0x0)

	CHECK_EQUAL_TEXTWAVES({str1}, {str2}, mode=WAVE_DATA)
End
```

gives

```
  Start of test "Unnamed"
  Entering test suite "EqualWaves_Examples.ipf"
  Entering test case "EqualWavesTextWithNulls"
  Assuming equality using mode WAVE_DATA for waves _free_ and _free_; detailed: Text waves difference:
Wave1:0x0857edab0
Wave2:0x0857ed3d0
Dimensions  |Labels|Text|
------------|------|----|
[0][0][0][0]|      |abcd|
            |      |abcd|

: is false.  Assertion "CHECK_EQUAL_TEXTWAVES({str1}, {str2}, mode=WAVE_DATA)" failed in line 16, procedure "EqualWaves_Examples.ipf"
  Leaving test case "EqualWavesTextWithNulls"
  Failed with 1 errors
  Leaving test suite "EqualWaves_Examples.ipf"
  Test finished with 1 errors
  End of test "Unnamed"
```

and makes the reader wonder what is different. Is there an easy way to solve that? Another column with the string lengths?

MichaelHuth:
Regarding the edgy case: The general issue is that strings can contain non-printable characters that can make a difference.
Printing the string length would add an indication but its not very reliable as the strings might even have the same length.
The only reliable way I would think of is:
If the string contains non-printable characters then add a hex code representation because the comparison itself is binary and not UTF-8 character based.

t-b:
Am Donnerstag, den 04.03.2021, 13:31 -0800 schrieb Michael Huth:

[...]

> The only reliable way I would think of is:
> If the string contains non-printable characters then add a hex code
> representation because the comparison itself is binary and not UTF-8
> character based.

Sounds like a good idea!

MichaelHuth:
ok,

regarding the last point in Code Issues: This was indeed a leftover from the state where the code had a separate part for INT64.
Though I noticed that my former local tests for INT64 were not correctly setup. As a result I had to reinclude parts for INT64 handling. Also the INT64 value printout was incorrect for unsigned and is still incorrect for complex values.
The problem with the latter is that only wfprintf supports printing complex wave elements but it converts to double before. This is reported, but I don't see a working option at the moment. Getting the component values by real() and imag() also converts to double.

t-b:
> The problem with the latter is that only wfprintf supports printing complex wave elements but it converts to double before. This is reported, but I don't see a working option at the moment. Getting the component values by real() and imag() also converts to double.

If IP does not have an option here, we need to live with it right? Just be sure to document the issue in our IP issue list.

t-b:
So the elements we can't printout properly are complex int64 wave elements? I'm fine with having that as a known limitation.

MichaelHuth:
In addition to the code issues the following is fixed:
- fixed incorrectly working comparison of 64 bit integer and complex 64 bit integer values (including tolerance).
- added UTF-8 character parser for extended output of strings

t-b:
-  [ ] Merge conflict
- [ ] Does that run without RTE on your side on latest IP9? Here I get a lot of RTEs.

MichaelHuth:
Also fixed the RTEs in IP6 and related possible RTEs in IP7/8 due to possible too long output strings with the EqualWaves table.

t-b (approved):
Very nice!

PR #130, opened 2020-05-20
Changed example8 to show that UTF takes care of RTE

### LINKED ISSUE #93: Update example 8 to adhere to the UTF evaluation of RTEs
Since the UTF evaluates runtime errors itself, it is in most typical usage scenarios not necessary to use AbortOnRTE in the test function.

This should be reflected in example 8, where e.g. an AbortOnValue can be caught without using AbortOnRTE.

MathiasAugstein:
First testcase got expanded to show how AbortOnRTE works without try-catch,
second testcase got changed to show that UTF takes care of the RTE.

In the manual an error was found. It was stated that AbortOnRTE gets treated as FAIL. FAIL aborts the execution of the test run, while AbortOnRTE only aborts the test case and gets later treated as CHECK (error counter and message).

Edit: Closes #93


docu/examples/example8-uncaught-runtime-errors.ipf
```diff
@@ -6,17 +6,23 @@
 
 Function TestWaveOp()
 
-		WAVE/Z/SDFR=$"I dont exist" wv;
+		WAVE/Z/SDFR=$"I dont exist" wv
+		print "This will be printed, even if a RTE occurs."
+
+		WAVE/Z/SDFR=$"I dont exist" wv; AbortOnRTE
```
MichaelHuth, line 12: If you want to show that the previous RTE is overwritten it would be good to have a different RTE condition before and having the related RTE error codes mentioned in the comment/description.

docu/examples/example8-uncaught-runtime-errors.ipf
```diff
@@ -6,17 +6,23 @@
...
 		WAVE/Z/SDFR=$"I dont exist" wv; AbortOnRTE
+		// If an RTE happens, the execution will jump to catch.
 		PASS()
 	catch
-		// Do not forget to clear the RTE
-		variable err = getRTError(1)
+		print "Here you can print additional info to understand the RTE."
+		// There is no need to clear the RTE (e.g. with GetRTError(1) )
+		// RunTest will take care and print the error message.
 		FAIL()
```
MichaelHuth, line 26: Add in the description that the FAIL prevents further execution after it and generates an additional error.

MathiasAugstein:
Also added the information, that the RTE in the second test case increases the error counter by 2 because of the the FAIL().

MichaelHuth:
I still don't agree with this part:

> :cpp:func:`PASS` increases
> the assertion counter and :cpp:func:`FAIL` treats this assertion as fail
> when a RTE was caught (i.e. increases the error counter and
> aborts the test run). That means the RTE in :code:`TestWaveOpSelfCatch`
> increases the error counter by 2.

How about:
In the case of the example runtime error the code continues at catch after AbortOnRTE triggered.
While the the runtime error condition is still active and the user reacts with a print the program runs to FAIL that aborts the test run with an error. The active runtime error condition is caught as well by the unit testing framework and generates another error. So TestWaveOpSelfCatch reports two errors, one from FAIL and one from the uncaught RTE.

I would not put much focus on the PASS as it is only included for completeness and not actually executed in the example. Also PASS and FAIL are separate assertions.

Issue #99, opened 2019-06-28
Increase error counter instead of silently ignoring missing generator functions


t-b:
The following example

```igorpro
function dostuff()

PASS()

end


Function MyTest([str])
	string str

End
```

results in

```console
•runtest("procedure", enableJu=1)
  Could not find data generator specification for multi data test case MyTest.
  Start of test "Unnamed"
  Could not find data generator specification for multi data test case MyTest.
  Entering test suite "Procedure"
  Entering test case "dostuff"
  Leaving test case "dostuff"
  Finished with no errors
  Leaving test suite "Procedure"
  Test finished with no errors
  End of test "Unnamed"
```

and

```xml
<?xml version="1.0" encoding="utf-8"?>
<testsuites>
	<testsuite package="Procedure" id="0" name="Procedure" timestamp="2019-06-28T08:50:00Z+02:00" hostname="THOMAS-WIN7-X64" tests="1" failures="0" errors="0" skipped="0" disabled="0" time="0.812">
		<properties>
			<property name="User" value="thomas"/>
			<property name="System" value="OS:Windows 7 Professional Service Pack 1;
OSVERSION:6.1.7601;
LOCALE:US;
IGORFILEVERSION:8.0.4.1"/>
			<property name="Experiment" value="HardwareTests"/>
			<property name="UTFversion" value="1.08"/>
			<property name="IgorInfo" value="IGORVERS:8.04;
BUILD:33900;
IGORKIND:pro64;
FREEMEM:8794046386176;
PHYSMEM:68656939008;
USEDPHYSMEM:1190862848;
NSCREENS:3;
SCREEN1:DEPTH=32,RECT=0,0,1200,1920;
SCREEN2:DEPTH=32,RECT=1200,0,2400,1920;
SCREEN3:DEPTH=32,RECT=2400,0,3600,1920"/>
		</properties>
		<testcase name="dostuff in Procedure (6)" classname="dostuff" time="0.804">
		<system-out>Entering test case &quot;dostuff&quot;
Leaving test case &quot;dostuff&quot;
</system-out>
		</testcase>
		<system-out>Entering test case &quot;dostuff&quot;
Leaving test case &quot;dostuff&quot;
</system-out>
	</testsuite>
</testsuites>
```

This hides the non-found generator message in the JUNIT output. I would have expected to see 2 testcases in JUNIT and one with an error/failure (not sure which one is appropriate). This should be also checked in related cases, we don't want to ignore tests silently.

PR #131, opened 2020-05-20
Missing data generators now create abort

MathiasAugstein:
If data generators for multi data tests are missing or somehow wrong,
the test run will be aborted and a descriptive message will be sent on
the command line.

A missing generator now results in an abort of the test run, in the same way as errors in the output of getTestCasesMatch get handled inside RunTest. If a data generator is not working, the whole test case is not working, which is similar to problems like not-finding-test-cases.

The descriptive messages, that were already coded, are now more highlighted and easy to read. Example:
`Error 64 in getTestCasesMatch: Could not find data generator specification for multi data test case GeneratorTest#MyTest.`

Closes: #99

### DIFF
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -1241,13 +1241,15 @@ static Function/S CheckFunctionSignaturesTC(testCaseList, procWin)
 	variable i, err, numTC, wType1, wType0
 	string fullTestCase, testCase, dgen, reducedTCList, msg
 
+	sprintf msg, "Error in test suite %s:\n \n", procWin
+
 	reducedTCList = ""
 	numTC = ItemsInList(testCaseList)
 	for(i = 0; i < numTC; i += 1)
 		testCase = StringFromList(i, testCaseList)
 		fullTestCase = getFullFunctionName(err, testCase, procWin)
 		if(err)
-			sprintf msg, "Could not get full function name of function: %s", fullTestCase
+			sprintf msg, "%sCould not get full function name of function: %s", msg, fullTestCase
 			UTF_PrintStatusMessage(msg)
 			continue
 		endif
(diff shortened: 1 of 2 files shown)


procedures/unit-testing-basics.ipf
```diff
@@ -2459,8 +2465,10 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
 		if(s.err)
 			sprintf msg, "Error %d in getTestCasesMatch: %s", s.err, s.allTestCasesList
 			UTF_PrintStatusMessage(msg)
-			sprintf msg, "Error: A test case matching the pattern \"%s\" could not be found in test suite(s) \"%s\".", s.testcase, s.procWinList
-			UTF_PrintStatusMessage(msg)
+			if(!(s.err & TC_MISSING_GENERATOR) )
```
MathiasAugstein, line 2468: Here a problem might occur, because in GetTestCasesMatch the line
`err = GetRTError(1)`
can also create an error with number 64.

MathiasAugstein, line 2468: Here a problem might occur, because in GetTestCasesMatch the line
`err = GetRTError(1)`
can also create an error with number 64.

MichaelHuth: In the following switch err is replaced by GREPLIST_ERROR

MathiasAugstein:
> Why is the propagation of the error up to RunTest needed?
You are completely right, this is not necessary. Will be changed into `Abort` inside `CheckFunctionSignaturesTC`.

MichaelHuth:
@MathiasAugstein I talked with Thomas and we decided to leave the last case as warning where the test run is continued. I will change that and also adapt the documentation.

MathiasAugstein:
Will be done once issue#9 is merged. Then this will be done by setting the expected_failure_flag.

t-b:
Please ping me once this is ready to merge, I want to test that with MIES and bamboo before merging.

MathiasAugstein:
just realized it is not trivial to mark a certain testcase as expected_failure or skip. My only thoughts about setting a testcase were hacky, I would need more time to solve this.
Right now they are just completely skipped, without being mentioned in either JUnit or TAP.

t-b:
Closing in favour of https://github.com/byte-physics/igor-unit-testing-framework/pull/159.

PR #137, opened 2020-05-29
Ma issue98 add check int64

### LINKED ISSUE #98: CHECK_EQUAL_INT64 and CHECK_EQUAL_UINT64 missing
Sometimes one could use an assertion for int64 data types as this is not covered by CHECK_EQUAL_VAR, where it is converted to double.

MathiasAugstein:
Added assertions to check if int64 variables are equal, close or small.

Default tolerance was set to 16 (4bit, as this should be close to the camera noise).

Closes #98


procedures/unit-testing-utils.ipf
```diff
@@ -58,7 +58,7 @@ threadsafe static Function IsInteger(var)
 	return trunc(var) == var && numtype(var) == 0
 End
 
-// @brief Convert a text wave to string list
+/// @brief Convert a text wave to string list
```
MichaelHuth, line 61: I know that this is a small fix, but please put unrelated fixes into a separate commit and PR.

procedures/unit-testing-utils.ipf
```diff
@@ -78,4 +78,70 @@ static Function/S TextWaveToList(txtWave, sep)
 	return list
 End
 
+#if IgorVersion() > 7.0
```
MichaelHuth, line 81: After our insight to int64/uint64 comparison logic in Igor, I think these functions are now obsolete?
If so please remove them.

t-b:
@MathiasAugstein Just a question from the sidelines. Can you point out a unit testing framework in other languages which has closeness/smallness defined for integers?

MathiasAugstein:
@t-b 
> @MathiasAugstein Just a question from the sidelines. Can you point out a unit testing framework in other languages which has closeness/smallness defined for integers?

I only found isAtLeast, isAtMost, isGreaterThan, isLessThan in the [truth assertion library](https://truth.dev/api/latest/index.html?overview-summary.html), which can also be used with JUnit. Should the assertion for smallness be replaced with IsAtMost, furthermore IsAtLeast added, and closeness be removed?

MichaelHuth:
@MathiasAugstein I added a fix that the documentation builds correctly and removed the pxp file from the VTTE commit. Effectivly the PR is good to merge now, but we want to delay that step until the CI issue is resolved.

PR #140, opened 2020-06-03
Ma issue111 search generators outside

### LINKED ISSUE #111: Multi data test cases should not require the generator function in the same procedure file
It should rather search the whole independent module/ProcGlobal.

MathiasAugstein:
Generator functions can now also be outside the procedure.
While testing I found out that static functions inside modules could already be addressed.
Manual was expanded to clarify how generators can be found.

Closes #111


MichaelHuth:
How about the following approach:

Apply FunctionInfo on funcNameReturn, you get the INDEPENDENTMODULE in the list by StringByKey.
Compare it to the current IM from GetIndependentModuleName() to decide if you are in the same or in another IM.

MathiasAugstein:
After finishing the changes I realized that it is impossible for `FunctionInfo` inside `GetDataGeneratorFunctionName` to find functions inside another independent module. You would need to set the correct procedure window with `procWin`, but that will always be the ipf of the test suite.

So the current state is: data generators inside the same independent module can be found globally, data generators inside other independent modules cannot be found, but the message will be, e.g.:
`Function TestModul#mygen in Independent Module ProcGlobal is unknown`

Should I remove the whole handling of possible independent module names as it is not used?

MichaelHuth (dismissed):

t-b:
Thanks.

Please add an experiment with run functions so that we can easily test that, now and in the future. I've added one, please have a look if this makes sense.

- I get

```
•runtest("test_example.ipf")
  Could not get full function name of data generator: In Independent Module ProcGlobal, data Generator Function TestExampleSub1#dataGenGlobal1 not found in test_example.ipf or globally in IM.
  Start of test "Unnamed"
  Could not get full function name of data generator: In Independent Module ProcGlobal, data Generator Function TestExampleSub1#dataGenGlobal1 not found in test_example.ipf or globally in IM.
  Entering test suite "test_example.ipf"
  Entering test case "TestExampleMain#MDTestCaseVar1:0"
  Leaving test case "TestExampleMain#MDTestCaseVar1:0"
  Entering test case "TestExampleMain#MDTestCaseVar2:0"
  Leaving test case "TestExampleMain#MDTestCaseVar2:0"
  Entering test case "TestExampleMain#MDTestCaseVar4:0"
  Leaving test case "TestExampleMain#MDTestCaseVar4:0"
  Finished with no errors
  Leaving test suite "test_example.ipf"
  Test finished with no errors
  End of test "Unnamed"
```

in IP9 but an abort message in IP8. There were some bugfixes in FunctionInfo/FunctionList in IP9.

Or is `dataGenGlobal1` a typo and it should be `dataGenGlobal`?

- The error message is also duplicated in IP9.

MichaelHuth:
@t-b ok, also in regard to https://github.com/byte-physics/igor-unit-testing-framework/pull/140 I think we should change the behavior to an Abort if the specified function could not be found.

t-b:
@MichaelHuth Sounds good.

MichaelHuth:
- rebased against main (instead of master), this resolves the Abort on fail as it was already in main.
- fixed typo "dataGenGlobal1"
- fixed test against IM. Interestingly FunctionInfo returns an empty value whereas GetIndependentModuleName() returns "ProcGlobal" when not running in an IM.

t-b:
Looks good now. Yes the default branch of all *public* repos has changed from master to main. The private ones are still todo.

t-b (approved):

PR #141, opened 2020-06-04
GetFullFunctionName bugfix to append module only when needed

### LINKED ISSUE #142: GetFullFunctionName appends module even if function is public
If the function is public no module name is required, so it should not be appended.

Furthermore if a function is inside a module, the return is simply the input of funcName. This was changed to the output desired in the commit 07086c5d606bef0fc90c89949bb2e321a253cb70.

Changing behavior for data generators and hooks must be tested.

MathiasAugstein:
Now the return only contains the module if the function is declared static.
Furthermore the code was cleaned up to apply better to our coding conventions.

Edit:
Tested with example5 (local and global hooks), output was exactly the same.
Tested with multiple data generators. Now UTF only finds data generators inside the same procedure file (as was intended), generator names like module#generator are not working (will be changed with issue#111).

Closes #142

### DIFF
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -757,8 +757,9 @@ static Function/S getFullFunctionName(err, funcName, procName)
 	string funcName, procName
 
 	err = FFNAME_OK
-	string infoStr = FunctionInfo(funcName, procName)
-	string errMsg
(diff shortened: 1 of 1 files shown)


MathiasAugstein:
Commit message was improved.

PR #144, opened 2020-06-05
Fix for wrong TC counting in TAP when SKIP was used with MD generators

MichaelHuth:
In a3ab7ca a bug was introduced that test cases were counted incosistently when using TAP.
Multi data tests that were SKIPed were counted globally correct, but in the output list only once.
This corrupted all subsequent test counting in TAP.

- When using a #TAPDirective: SKIP on Multi Data Tests the global TAP count
  included the MD data tests but the skipped MD test was only counted once.
- Moved the TAP skip decision to the execution block for the TC function
  and TC hooks
- Removed unneeded call to TAP_GetNotes after each skipped TC

Bug was introduced in a3ab7ca


t-b:
@MichaelHuth Is there a experiment in internal_dev I can run to see the difference?

MichaelHuth:
Yes, try with 
docu/examples/TAP_Examples/tap_example1a-uncaught-aborts.ipf

t-b:
@MichaelHuth The cited example does not have a multidata test case.

I came up with

```
// This is the first data generator function used by MDTestCaseVar
static Function/WAVE tcDataGenVar()
	Make/FREE data = {5, 1}
	SetDimLabel 0, 0, first, data
	SetDimLabel 0, 1, second, data
	return data
End

// #TAPDirective: SKIP with TAP enabled this Test Case is not executed (skipped) and reported 'ok', when executed it fails
// UTF_TD_GENERATOR tcDataGenVar
Function TAPTestSkip()
	CHECK_EQUAL_VAR(1,2)
End
```

which gives both on master and here

```
TAP version 13
1..0 All test cases marked SKIP
```

What am I doing wrong?

MichaelHuth:
@t-b Looks fully correct. TAP has a special output if all testcases are marked SKIP.
If you add one non-skip test case it should output the TC list.

t-b:
Using

```
#pragma rtGlobals=3		// Use modern global access method and strict wave access.
#include "unit-testing"

// This is the first data generator function used by MDTestCaseVar
static Function/WAVE tcDataGenVar()
	Make/FREE data = {5, 1}
	SetDimLabel 0, 0, first, data
	SetDimLabel 0, 1, second, data
	return data
End

// #TAPDirective: SKIP with TAP enabled this Test Case is not executed (skipped) and reported 'ok', when executed it fails
// UTF_TD_GENERATOR tcDataGenVar
Function TAPTestSkip()
	CHECK_EQUAL_VAR(1,2)
End

Function Test()

	PASS()
End
```

I'm getting both in master and here 

```
TAP version 13
1..2
ok 1  # SKIP with TAP enabled this Test Case is not executed (skipped) and reported 'ok', when executed it fails
ok 2
```

so I don't see a change with this PR. Any ideas?

PR #145, opened 2020-06-10
Bugfix tap examples

t-b:
Please merge if good.


MichaelHuth:
@t-b I think a better fix would be to replace the AbortNow() by FAIL().
Rest is good.

t-b:
hmm okay!

PR #159, opened 2020-10-10
CheckFunctionSignaturesTC: Abort on invalid data generator functions

### LINKED ISSUE #99: Increase error counter instead of silently ignoring missing generator functions

t-b:
Ever since the introduction of multi data test cases we only warned when
the data generator function could not be used.

This hides buggy test setups and is there bad.

So let's abort in these cases.

The case of an empty generator wave will be handled separately.

Close #99 

@MichaelHuth I went for the abort solution you suggested in https://github.com/byte-physics/igor-unit-testing-framework/pull/131#pullrequestreview-415353210.

### DIFF
--- a/docu/sphinx/source/advanced.rst
+++ b/docu/sphinx/source/advanced.rst
@@ -286,7 +286,7 @@ The data generator function name must be attributed with a comment within four
 lines above the test cases Function line. The key word is `UTF_TD_GENERATOR` with
 the data generators function name following as seen in the simple example here.
 If no data generator is given or the format of the test case function does not fit
-to the wave type then a error message is printed and the test case is ignored.
+to the wave type then a error message is printed and the test run is aborted.
 
 The test case names are by default extended with `:num` where num is the index
 of the wave returned from the data generator. For convenience in the data generator
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -1356,33 +1356,33 @@ static Function/S CheckFunctionSignaturesTC(testCaseList, procWin)
 		if(err)
 			sprintf msg, "Could not find data generator specification for multi data test case %s. %s", fullTestCase, dgen
 			UTF_PrintStatusMessage(msg)
-			continue
+			Abort msg
 		else
 			dgen = getFullFunctionName(err, dgen, procWin)
 			if(err)
 				sprintf msg, "Could not get full function name of data generator: %s", dgen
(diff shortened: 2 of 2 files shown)


MichaelHuth (approved):
Looks good.

PR #163, opened 2020-11-21
Avoid bugging out on long strings due to snprintf limitation

### LINKED ISSUE #76: CHECK_EQUAL_STR: fails for very long strings
Again we hit the sprintf limit with

```
Function mytest()

	string a = PadString("a", 10e4, 0x20)
	string b = PadString("b", 10e4, 0x20)
	CHECK_EQUAL_STR(a, b)
	CHECK_PROPER_STR(a)
End
```

as that gives

```
•runtest("procedure")
  Start of test "Unnamed"
  Entering test case "mytest"
  " not respecting case: is false
  Assertion "CHECK_EQUAL_STR(a, b)" failed in line 9, procedure "Procedure"
  Uncaught runtime error 2006:"printf;string too long" in test case "mytest", procedure file "Procedure"
  Leaving test case "mytest"
  Failed with 2 errors
  Test finished with 2 errors
  End of test "Unnamed"
```

t-b:
Close #76.


procedures/unit-testing-utils.ipf
```diff
@@ -188,4 +188,30 @@ static Function/T GetFunctionTagValue(funcName, tagName, err)
...
+/// @brief Prepare the passed string for output
+///
+/// We return a fixed string if it is null and limit its size so that it can be used in a sprintf statement using plain
+/// "%s". We also do that for IP9, where this limit does not exist anymore to have a consistent output across IP
+/// versions.
+static Function/S PrepareStringForOut(str)
+	string &str
+
+	variable length
+	variable maxLength = 250
```
MichaelHuth, line 201: How about setting this and the suffix as static constant?

MichaelHuth (changes requested):
Also tested and works. 

The additional test is good to prevent accidental introduction of some output later that violates the length limitation. 
👍

procedures/unit-testing-utils.ipf
t-b: @MichaelHuth Thx. Fixed. I also explained how I reached that value.

MichaelHuth (approved):

PR #173, opened 2021-02-24
docu/sphinx/source: Always use tab-width of four spaces

### LINKED ISSUE #171: Use :tab-width: 4 for all code exmaples

t-b:
This is the standard for IP code.

Close #171.

### DIFF
--- a/docu/sphinx/source/examples.rst
+++ b/docu/sphinx/source/examples.rst
@@ -24,6 +24,7 @@ function is also capable of handling :code:`INF` singularities.
 
 .. literalinclude:: ../../examples/example1-plain.ipf
    :caption: example1-plain.ipf
+   :tab-width: 4
    :linenos:
    :emphasize-lines: 11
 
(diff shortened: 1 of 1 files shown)


docu/sphinx/source/examples.rst
```diff
@@ -24,13 +24,15 @@ function is also capable of handling :code:`INF` singularities.
...
    :caption: example1-plain.ipf
+   :tab-width: 4
    :linenos:
    :emphasize-lines: 11
 
 The test suite can be executed using the following command:
 
 .. code-block:: igor
    :caption: command
+   :tab-width: 4
```
MichaelHuth, line 35: The :tab-width:  option does not exist for .. code-block blocks.
These blocks have to be changed to literalinclude blocks for that.

MichaelHuth (changes requested):
tab-width does not work with code-blocks.

docu/sphinx/source/examples.rst
t-b: @MichaelHuth Good point. I've removed them from all code-blocks. Thinking about it is also does not make sense as we in these.

MichaelHuth (approved):

Issue #175, opened 2021-02-25
Generator functions most now be static inside modules


t-b:
Latest version d5c7d320 (Merge pull request #174 from byte-physics/speedup-build, 2021-02-24) gives on the following example

```igorpro
#pragma rtGlobals=3
#pragma ModuleName=Example13

#include "unit-testing"

Function/WAVE tcDataGenVar()
	Make/FREE data = {5, 1}
	SetDimLabel 0, 0, first, data
	SetDimLabel 0, 1, second, data
	return data
End

// UTF_TD_GENERATOR tcDataGenVar
Function MDTestCaseVar([var])
	variable var

	CHECK(var == 1 || var == 5)
End
```

the abort `Could not get full function name of data generator: Data Generator Function Example13#tcDataGenVar must be static.` whereas it works in c5c37a9f (Merge pull request #169 from byte-physics/fix-sphinx-warning, 2021-02-17).

This is due to #140. Is that intentional? It seems we now enforce that it must be static when it is inside a regular module. But shouldn't it be the otherway around, i.e. if it is static it must be in a regular module?

I found that with RunAllTests.pxp in MIES.

t-b:
Just to be clear. We open to fix the tests if the current behaviour makes more sense. I just need to understand it first.

MichaelHuth:
Yes, in https://github.com/byte-physics/igor-unit-testing-framework/pull/140 it is enforced. Though I think the better logic would be as you wrote that if it is static then it must be in a regular module.

The implications of the current enforcing is that either one changes the generator to static or moves all global generators to an own ipf file. This is stronger than necessary. I will change it.

PR #176, opened 2021-02-25
Fix: Do not enforce Datagenerator function to be static in same procedure.

MichaelHuth:
In commit 98f58595e5dfd130d7b1d8e44f728724e70bca6e it was introduced that the
data generator function was enforced to be static when it was present in the
same procedure file.

However when the function is static then the procedure must have a Module
specification. Thus, the logic was changed that it gives an error only if
a static function was found without Module specification.
Also without Module specification (for non-static functions) only the function
name is returned.

Also note that FunctionInfo does not find static functions in other procedure files when called with the function name specification only. So the check is not necessary for the else part.

close https://github.com/byte-physics/igor-unit-testing-framework/issues/175


t-b:
@MichaelHuth Thx. Works now.

Issue #113, opened 2019-12-19
Save and restore assertion counters before and after calling user hooks


t-b:
The following example

```igorpro
#include "unit-testing"

Function TEST_CASE_BEGIN_OVERRIDE(string str)

	PASS()
End

Function mytest()

End
```

produces

```
•runtest("procedure", testcase="mytest")
  Start of test "Unnamed"
  Entering test suite "Procedure"
  Entering test case "mytest"
  Leaving test case "mytest"
  Finished with no errors
  Leaving test suite "Procedure"
  Test finished with no errors
  End of test "Unnamed"
```

which is unexpected as there should be a warning about not having any assertions in `mytest`. It looks like we only need to save and restore the assertion count in `ExecuteHooks` prior to calling the user defined hook.

PR #177, opened 2021-03-02
Enforce assertion in test case, not counting assertions from test hooks

MichaelHuth:
Only assertions in the test case are counted when enforcing at least one assertion per test case.
A test case spans all code after test case begin to test case end including possible reentry functions.

close https://github.com/byte-physics/igor-unit-testing-framework/issues/113

### DIFF
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -610,6 +610,38 @@ static Function IsProcGlobal()
 	return !cmpstr("ProcGlobal", GetIndependentModuleName())
 End
 
+/// @brief Saves the current assertion counter in global saveAssertCount
+static Function SaveAssertionCounter()
+
+	DFREF dfr = GetPackageFolder()
+
+	NVAR/SDFR=dfr/Z assert_count
+	if(!NVAR_Exists(assert_count))
+		initAssertCount()
+		NVAR/SDFR=dfr assert_count
+	endif
+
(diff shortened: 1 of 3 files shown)


procedures/unit-testing-basics.ipf
```diff
@@ -2201,6 +2233,7 @@ static Function CallTestCase(s, reentry)
 		UTF_PrintStatusMessage(msg)
 	else
 		func = s.fullFuncName
+		SaveAssertionCounter()
```
t-b, line 2236: Is there  a reason we don't call that in TestCaseBegin? This would be symmetrical compared to the call to GetSavedAssertionCounter.

t-b (changes requested):

procedures/unit-testing-basics.ipf
MichaelHuth: The reason is that it is not symmetrical. By our definition the user hook for test case begin is called after TestCaseBegin in ExecuteHooks.

t-b:
@MichaelHuth With the example you provided I'm getting

```
Test case "myTestCase" contained at least one assertion: is false.  Assertion failed in unknown location
```

When I comment out both `PASS()` statements I get the expected

```
Test case "myTestCase" contained at least one assertion: is false.  The test case did not make any assertions!
```

MichaelHuth:
Ok, so the evaluation of "internal" assertion need to be adapted.

t-b (approved):

Issue #134, opened 2020-05-22
Document JUNIT standard implemented


t-b:
For TAP we already document that at https://docs.byte-physics.de/igor-unit-testing-framework/advanced.html#test-anything-protocol-output.

We should do that for JUNIT as well. Incorporate the XSD we used here in the repo (or even docu?).

The website I used is https://llg.cubic.org/docs/junit:  [Website dump](https://github.com/byte-physics/igor-unit-testing-framework/files/4669616/JUNIT.website.dump.zip)

@MichaelHuth Does that reference look good to you?

PR #178, opened 2021-03-02
Added JUNIT reference and updated JUNIT section in advanced.rst file

MichaelHuth:
- Added example XML and XSD definition sheet for the JUNIT implementation that
  the IUTF uses.
- Added links to official external sources of at the Jenkins xunit
  implementation on github.

close https://github.com/byte-physics/igor-unit-testing-framework/issues/134

### DIFF
--- a/docu/sphinx/source/advanced.rst
+++ b/docu/sphinx/source/advanced.rst
@@ -83,17 +83,18 @@ cleanup is done afterward. The *next* Test Case then starts with the data the
 JUNIT Output
 ------------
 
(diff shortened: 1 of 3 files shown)


MichaelHuth:
I have forced the xml parser to parse the example xml file as it errored out that the lexer is unable to parse it.
It has some syntax coloring now, but the result is rather mixed. Although I am sure it is a valid xml file.

t-b:
Jeep the XML/XSD seems to be too complicated for the parser. I'll still merge as is as that is nothing we can fix now.

t-b (approved):

Issue #112, opened 2019-11-24
multi data test case: Allow more function signatures with wave arguments


t-b:
The following test case

```igorpro
Function/WAVE Gen()

	Make/FREE/WAVE wv
	return wv
End

// UTF_TD_GENERATOR Gen
Function TestCase([WAVE/WAVE wv])
End
```

is not considered a valid test case as its function signature does not match one of the existing ones in https://github.com/byte-physics/igor-unit-testing-framework/blob/e1f316036131e13a9634c2a2ba5c37479f42b622/procedures/unit-testing-basics.ipf#L1177-L1220.

This requires users to work around by creating another wave reference. It would be nice to support these signatures out of the box.

MichaelHuth:
There is already a generic wave type function template implemented and the correct code would be:

```
Function/WAVE Gen()

	Make/FREE/WAVE wv
	return wv
End

// UTF_TD_GENERATOR Gen
Function TestCase([WAVE wv])

	WAVE/WAVE w = wv
End

```

Do you think we should extend that to support directly typed waves as function parameter?
like:
WAVE/T
WAVE/DF
WAVE/WAVE
?

PR #181, opened 2021-03-04
Introduced typed waves for multi data test cases

MichaelHuth:
In addition to the generic WAVE type multi data test cases also accept now
WAVE/T, WAVE/WAVE and WAVE/DF specific types.
The associated data generators must return waveRef waves where each element
must have the fitting type.

Added examples for all three new typed waves to DataGenSpecificWaveTypes.ipf
in internal_dev.

The MD test case data type setup is not internally buffered for each test case
found. This means that for the actual call to the test case the function
template is determined from the wave type of the element of the data generator
wave.
So for a data generator that returns a wave reference wave with text waves as
elements a test case with WAVE and with WAVE/T is valid. This applies to
reentry functions as well. See also the last example in
DataGenSpecificWaveTypes.ipf.
There is no priority applied as with a test case name only one declaration is
connected.

So the test cases entry function declaration puts constraints on the wave type of the elements from the wave reference wave from the data generator. Reentry functions are checked at runtime because they are also known through the registration. For these WAVE and the typed WAVE/xxx declaration might be valid, then both is checked and the call is done according to the found function declaration. Thus a test case entry when declared with WAVE/T constrains all elements from the data generator wave reference wave to text wave type. A following reentry function is valid with either WAVE or WAVE/T.

close https://github.com/byte-physics/igor-unit-testing-framework/issues/112

### DIFF
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -931,6 +931,36 @@ Function TEST_CASE_PROTO_MD_WV([wv])
 	abortNow()
 End
 
+Function TEST_CASE_PROTO_MD_WVTEXT([wv])
+	WAVE/T wv
+
+	string msg
+
+	sprintf msg, "Error: Prototype function %s was called.", GetRTStackInfo(1)
+	UTF_PrintStatusMessage(msg)
+	abortNow()
+End
+
+Function TEST_CASE_PROTO_MD_WVDFREF([wv])
+	WAVE/DF wv
+
+	string msg
+
+	sprintf msg, "Error: Prototype function %s was called.", GetRTStackInfo(1)
+	UTF_PrintStatusMessage(msg)
+	abortNow()
+End
+
+Function TEST_CASE_PROTO_MD_WVWAVEREF([wv])
+	WAVE/WAVE wv
+
+	string msg
+
+	sprintf msg, "Error: Prototype function %s was called.", GetRTStackInfo(1)
+	UTF_PrintStatusMessage(msg)
+	abortNow()
+End
+
 Function TEST_CASE_PROTO_MD_DFR([dfr])
 	DFREF dfr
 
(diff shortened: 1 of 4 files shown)


procedures/unit-testing-basics.ipf
```diff
@@ -1440,9 +1484,20 @@ static Function/S CheckFunctionSignaturesTC(testCaseList, procWin)
...
+			elseif(!UTF_Utils#IsNaN(wRefSubType) && wType1 == WAVETYPE1_WREF)
+				WAVE/WAVE wWrefType = wGenerator
+				numMD = DimSize(wGenerator, UTF_ROW)
+				for(i1 = 0; i1 < numMD; i1 += 1)
+					WAVE wCheckType = wWrefType[i1]
+					if(WaveType(wCheckType, 1) != wRefSubType)
+						sprintf msg, "Test case %s expects specific wave type1 %u from the Data Generator %s. The wave type at index %u from the data generator does not fit to expected wave type.", fullTestCase, wRefSubType, dgen, i
+						UTF_PrintStatusMessage(msg)
+						Abort msg
+					endif
```
t-b, line 1496: How about a sub routine here like

```
static Function HasConstantWaveTypes(wv, subType)
WAVE/WAVE wv
variable subType

Make/FREE/N=(DimSize(wv, UTF_ROWS)) matches = WaveType(wv[0], 1) == wRefSubType

FindValue/V=(0) matches
return V_Value == -1
```

and then

```
if(!HasConstantWaveTypes(wWrefType, wRefSubType))
```

procedures/unit-testing-basics.ipf
```diff
@@ -931,6 +931,36 @@ Function TEST_CASE_PROTO_MD_WV([wv])
 	abortNow()
 End
 
+Function TEST_CASE_PROTO_MD_WVTEXT([textWave])
```
t-b, line 934: Can we use `wv` as optioal parmeter name as well? This makes writing test cases which use that feature easier as I don't have to remember the specialized name.

t-b (changes requested):

procedures/unit-testing-basics.ipf
MichaelHuth: ok, utility function is added and used.

procedures/unit-testing-basics.ipf
MichaelHuth: Changed optional parameters to wv as well.
Though the name in the test case/reentry functions do not matter as it is an optional parameter.

t-b (approved):
LGTM.

PR #182, opened 2021-03-05
Fix for RTE cleanup in AreWavesEqual blocking TC RTE evaluation

MichaelHuth:
A fix for invalid parameters for a EqualWaves call in the IUTF was introduced
in 28f7fc37221c117ae3d4a6e5baddd4dd549cc4bb

This created the problem that an RTE that happens in a test case is cleared
at this point in the IUTF and then the RTE gets not visible in the RTE
evaluation. Thus it gets ignored and does not raise an error.

The fix:
EqualWaves is still "monitored" with a AbortOnRTE but not within a try catch
block. If there is a lingering RTE or EqualWaves generates an RTE then the
catch block for the test case takes action and evaluates the RTE.
Thus, wrong parameter passed from CHECK_EQUAL_WAVES to IUTF EqualWaves raise
an RTE that looks like it originates from the test case itself. This is valid
because the CHECK_EQUAL_WAVES parameters are compatible with EqualWaves.

Also mode is checked in EQUAL_WAVE_WRAPPER for validity to abort early on
invalid modes. mode must have at least one valid bit set.

### DIFF
--- a/procedures/unit-testing-assertion-checks.ipf
+++ b/procedures/unit-testing-assertion-checks.ipf
@@ -200,13 +200,7 @@ static Function AreWavesEqual(wv1, wv2, mode, tol, detailedMsg)
 
 	variable result, err
 
-	// handle NaN return values from EqualWaves for unknown modes
-	try
-		result = EqualWaves(wv1, wv2, mode, tol) == 1; AbortOnRTE
-	catch
-		err = GetRTError(1)
-		result = 0
-	endtry
+	result = EqualWaves(wv1, wv2, mode, tol) == 1
 
 	detailedMsg = ""
 
--- a/procedures/unit-testing-assertion-wrappers.ipf
+++ b/procedures/unit-testing-assertion-wrappers.ipf
@@ -667,11 +667,25 @@ static Function EQUAL_WAVE_WRAPPER(wv1, wv2, flags, [mode, tol])
 	if(ParamIsDefault(mode))
 		Make/I/FREE modes = { WAVE_DATA, WAVE_DATA_TYPE, WAVE_SCALING, DATA_UNITS, DIMENSION_UNITS, DIMENSION_LABELS, WAVE_NOTE, WAVE_LOCK_STATE, DATA_FULL_SCALE, DIMENSION_SIZES}
 	else
+		if(!UTF_Utils#IsFinite(mode))
+			EvaluateResults(0, "Valid mode for EQUAL_WAVE check.", flags)
+			return NaN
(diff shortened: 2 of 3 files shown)


MichaelHuth:
I realized that the approach is not compatible with IP6. I will realize another patch.

MichaelHuth:
I have removed the whole AbortOnRTE approach and do the check manually before now.
The issue with that is that IP6 does not raise an RTE at all, with wrong parameters. There EqualWaves just returns NaN.

t-b:
Thanks, 

I've applied the following minor cleanup

```diff
diff --git a/procedures/unit-testing-assertion-wrappers.ipf b/procedures/unit-testing-assertion-wrappers.ipf
index 6426d07..038b7d0 100644
--- a/procedures/unit-testing-assertion-wrappers.ipf
+++ b/procedures/unit-testing-assertion-wrappers.ipf
@@ -670,22 +670,22 @@ static Function EQUAL_WAVE_WRAPPER(wv1, wv2, flags, [mode, tol])
         if(!UTF_Utils#IsFinite(mode))
             EvaluateResults(0, "Valid mode for EQUAL_WAVE check.", flags)
             return NaN
-        endif
-        if(!(mode & (WAVE_DATA | WAVE_DATA_TYPE | WAVE_SCALING | DATA_UNITS | DIMENSION_UNITS | DIMENSION_LABELS | WAVE_NOTE | WAVE_LOCK_STATE | DATA_FULL_SCALE | DIMENSION_SIZES)))
+        elseif(!(mode & (WAVE_DATA | WAVE_DATA_TYPE | WAVE_SCALING | DATA_UNITS | DIMENSION_UNITS | DIMENSION_LABELS | WAVE_NOTE | WAVE_LOCK_STATE | DATA_FULL_SCALE | DIMENSION_SIZES)))
             EvaluateResults(0, "Valid mode for EQUAL_WAVE check.", flags)
             return NaN
         endif
+
         Make/I/FREE modes = { mode }
     endif

     if(ParamIsDefault(tol))
         tol = 0.0
     elseif(UTF_Utils#IsNaN(tol))
-            EvaluateResults(0, "Valid tolerance for EQUAL_WAVE check.", flags)
-            return NaN
+        EvaluateResults(0, "Valid tolerance for EQUAL_WAVE check.", flags)
+        return NaN
     elseif(tol < 0)
-            EvaluateResults(0, "Valid tolerance for EQUAL_WAVE check.", flags)
-            return NaN
+        EvaluateResults(0, "Valid tolerance for EQUAL_WAVE check.", flags)
+        return NaN
     endif

     for(i = 0; i < DimSize(modes, 0); i += 1)
```

and will merge then CI passes.

t-b (approved):

Issue #26, opened 2018-10-12
JUNIT: Emit multiple failure tags for multiple failed assertions


t-b:
```igorpro
#include "unit-testing"

Function Test()

	CHECK_EQUAL_VAR(1, 2)
	CHECK_EQUAL_VAR(2, 3)
	CHECK_EQUAL_VAR(3, 4)	
End
```

run with `Runtest("Procedure", enableJU=1)`

gives

```xml
<?xml version="1.0" encoding="utf-8"?>
<testsuites>
        <testsuite package="Procedure" id="0" name="Procedure" timestamp="2021-03-03T14:31:05" hostname="WIN10-THOMAS-PC" tests="1" failures="1" errors="0" skipped="0" disabled="0" time="0.009">
                <properties>
                        <property name="User" value="thomas"/>
IGORFILEVERSION:9.0.0.4"/>roperty name="System" value="OS:Windows 10 Pro;
                        <property name="Experiment" value="Experiment"/>
                        <property name="UTFversion" value="1.08"/>
SCREEN3:DEPTH=32,RECT=1200,0,2400,1920"/>gorInfo" value="IGORVERS:9.00;
                </properties>
                <testcase name="Test in Procedure (1)" classname="Test" time="0.005">
                        <failure message="3 == 4: is false.  Assertion &quot;CHECK_EQUAL_VAR(3, 4)&quot; failed in line 10, procedure &quot;Procedure&quot;" type="FAIL"></failure>
</system-out>case &quot;Test&quot;;CHECK_EQUAL_VAR(3, 4)&quot; failed in line 10, procedure &quot;Procedure&quot;
                <system-err>3 == 4: is false.  Assertion &quot;CHECK_EQUAL_VAR(3, 4)&quot; failed in line 10, procedure &quot;Procedure&quot;</system-err>
                </testcase>
</system-out>case &quot;Test&quot;;CHECK_EQUAL_VAR(3, 4)&quot; failed in line 10, procedure &quot;Procedure&quot;
                <system-err>3 == 4: is false.  Assertion &quot;CHECK_EQUAL_VAR(3, 4)&quot; failed in line 10, procedure &quot;Procedure&quot;</system-err>
        </testsuite>
</testsuites>
```

Only the last assertion made it into the file. This might be a regression.

Desired outcome:
- One failure tag per failed assertion
- correct failure count in testsuite tag as well

[...]

PR #183, opened 2021-03-05
With JUnit enabled a <failure> tag for each failed assertion is written

MichaelHuth:
Previously only the last failure message appeared in the JUnit output.
Now for each failed assertion a <failure> or <error> tag is written,
depending if the test case aborted or not.

Also <system-err> output was extended accordingly to include all
fail messages.

The xml file with multiple failure tags was checked against the
reference xsd schema file and it validates (using xmllint)

close https://github.com/byte-physics/igor-unit-testing-framework/issues/26

### DIFF
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -462,6 +462,8 @@ Function incrError()
 		NVAR/SDFR=dfr error_count
 	endif
 
+	AddMessageToBuffer()
+
 	error_count +=1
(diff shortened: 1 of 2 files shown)


t-b:
Gitlab:

grafik

grafik

Bamboo:
TODO

t-b (approved):

MichaelHuth:
I think thats ok although Bamboo seems to count each sub failure as well.

PR #187, opened 2021-03-09
fix for IUTF not compiling in Independent Module

MichaelHuth:
includes a new experiment that tests if it compiles in an IM with an empty passing test.


t-b:
I know I'm not assigned but you want to commit a non-empty pxp.

t-b:
@MichaelHuth Sorry for the noise. Githubs UI says so. But locally it is not empty.

Issue #28, opened 2018-11-21
Investigate possibilities to get code coverage information


t-b:
See http://www.igorexchange.com/node/7918 for why the easy way is not possible.

Instrumenting the code before a la

```
Function Track(file, line)
    string file
    variable line

    return 1
End

Function MyTest()
    Track("myprocedure", 123); 
    Track("myprocedure", 124); variable a
    if(Track("myprocedure", 125) && a)
        Track("myprocedure", 126); a += 1
    else
        Track("myprocedure", 127); a += 2
    endif
    Track("myprocedure", 128); 
End
```

seems doable. But flow controls statements and line continuation need special handling.

Output format from our UTF could be the one from cobertura (XML based), see https://leonard.io/blog/2011/06/cobertura-sample-xml-output/ and http://cobertura.sourceforge.net/xml/coverage-03.dtd.

Gitlab can not yet parse that https://gitlab.com/gitlab-org/gitlab-ce/issues/44406 but this is planned.

MichaelHuth:
**COVXOP_Init number:maxProcs number:maxLines**
Creates and holds a wave reference wave with a reserved start size (e.g. 1000) with null refs.
Stores maxProcs, maxLines

**COVXOP_Trace number:procNum number:lineNum**
These parameters are generated when the Z_(p, l) cal is inserted.
Idea:
We have a vector that maps thread Ids to indices into the wave reference wave.
Get thread id: currentThreadID = (PSInt)GetCurrentThreadId()
Use std::find to find it in vector.
When not found:
{
mutex
add new threadid to vector
if required increase wref wave size by e.g. another 1000
create new wave with maxProcs, maxLines size and put in wref[index]
}
w = wref[index]
w[procNum][lineNum] += 1

**COVXOP_GetResults DataFolderAndName:{dest, waveref}**
Returns the wave reference wave
We might combine this with releasing the hold by the XOP and drop COVXOP_Release

**COVXOP_Release**
Releases the hold of the wave ref wave from the XOP.

t-b:
I'd imagine the XOP having a more generic interface. One which we could reuse for other projects as well.

Why does COVXOP_Trace not just return a wave? And similiarly we could also just put a free wave reference wave into the XOP with COVXOP_Init or?

MichaelHuth:
Rev2:
@t-b

**COVXOP_Init wave:waveIn**
waveIn is a wave reference wave that the XOP holds.

[...]

PR #192, opened 2021-04-01
WIP: Generate code coverage data

MichaelHuth:
How to play:

Needs: 
IP9 beta
Create in C:\ a folder "trace"

For awareness of static functions, the ipf files require to have a module name.
e.g. for MIES #define AUTOMATED_TESTING
that enables Module Names for most ipfs.

Call:
SetupTraceProcedures("MIES.*.ipf",1)
First param is a procedure list, second param says if first param is a regular exp or not.

The *.ipf files get replaced in place.

Compile and then logging is active.


t-b:
I've played around with it. It's very cool that the instrumentation already works with MIES. I did not expect that so quickly. As you have already mentioned the execution speed is quite slow. And in the case of MIES I would even say unbearable slow.

I came up with the following example to play around

```igorpro
#pragma rtGlobals=3				// Use modern global access method and strict wave access
#pragma DefaultTab={3,20,4}		// Set default tab width in Igor Pro 9 and later

//threadsafe static Function Z_()
//
//End

Function Run_IGNORE()
	variable now = stopmstimer(-2)
	RunTest("Proc0", testcase = "Testcase_Single")
	printf "%g\r", (stopmstimer(-2) - now)/1e6	

	now = stopmstimer(-2)
	RunTest("Proc0", testcase = "Testcase_Multi")
	printf "%g\r", (stopmstimer(-2) - now)/1e6	
End

// Tests

Function Testcase_Single()

	CHECK_EQUAL_VAR(MyCoolCode(0), 0)
End

Function Testcase_Multi()

	CHECK_EQUAL_VAR(MyCoolCode(1), 1)
End

// Manually instrumented code

threadsafe Function UberCool(variable val)

	Z_(); return val
End

// Uses multithread statement if param == 1, single thread otherwise
Function MyCoolCode(variable param)

Z_();	Make/FREE/N=(1024) wv

	if(param == 1)
Z_();		Multithread wv = Ubercool(p)
Z_();		print "hi there"
Z_(); 		return 1
	else
Z_();		wv = Ubercool(p)
Z_();		print "not here"
Z_(); 		return 0
	endif
End
```

Normal run:
```
•run_IGNORE()
  Start of test "Unnamed"
  Entering test suite "Proc0"
  Entering test case "Testcase_Single"
  not here
  Leaving test case "Testcase_Single"
  Finished with no errors
  Leaving test suite "Proc0"
  Test finished with no errors
  End of test "Unnamed"
  2.51508
  Start of test "Unnamed"
  Entering test suite "Proc0"
  Entering test case "Testcase_Multi"
  hi there
  Leaving test case "Testcase_Multi"
  Finished with no errors
  Leaving test suite "Proc0"
  Test finished with no errors
  End of test "Unnamed"
  2.91445
```

Note that execution *with* multithread takes longer as single thread.

Commenting in `Z_()` which simulates non-instrumented code results in
```
•run_IGNORE()
  Start of test "Unnamed"
  Entering test suite "Proc0"
  Entering test case "Testcase_Single"
  not here
  Leaving test case "Testcase_Single"
  Finished with no errors
  Leaving test suite "Proc0"
  Test finished with no errors
  End of test "Unnamed"
  0.0588005
  Start of test "Unnamed"
  Entering test suite "Proc0"
  Entering test case "Testcase_Multi"
  hi there
  Leaving test case "Testcase_Multi"
  Finished with no errors
  Leaving test suite "Proc0"
  Test finished with no errors
  End of test "Unnamed"
  0.0435093
```

So it is around 50 times slower.

So the question is how can we make Z_() faster.

t-b:
If I reduce `Z_()` to do only Open/Close I get around 1.8/2.6s.

t-b:
I'm attaching a benchmark from https://github.com/cameron314/concurrentqueue, which I've produced on my machine. Different queues, compiled with g++.

[benchmarks.log](https://github.com/byte-physics/igor-unit-testing-framework/files/6270494/benchmarks.log)

The relevant benchmark for us is IMHO

```
only enqueue (pre-allocated):
  (Measures the average operation speed when all threads are producers,
  and the queue has been stretched out first)
  > moodycamel::ConcurrentQueue
      Without tokens
         1    thread:  Avg: 0.0110us  Range: [0.0105us, 0.0114us]  Ops/s:  90.53M  Ops/s/t:  90.53M
         2   threads:  Avg: 0.0525us  Range: [0.0300us, 0.0687us]  Ops/s:  38.13M  Ops/s/t:  19.06M
         4   threads:  Avg: 0.1002us  Range: [0.0729us, 0.1843us]  Ops/s:  39.90M  Ops/s/t:   9.98M
         8   threads:  Avg: 0.2294us  Range: [0.2191us, 0.2566us]  Ops/s:  34.87M  Ops/s/t:   4.36M
         32  threads:  Avg: 1.2175us  Range: [1.1860us, 1.2405us]  Ops/s:  26.28M  Ops/s/t: 821.38k
         Operations per second per thread (weighted average):  11.97M
```

as we should be able to preallocate. With some bit-shifting we should be able to fit line and file index in an integer, 32bit would work.

t-b:
Example playing around: 
[Experiment.zip](https://github.com/byte-physics/igor-unit-testing-framework/files/6273370/Experiment.zip)

t-b:
Attached is an very-hacky visualisation of the coverage data.
[Notebook-Tracing-Chirp.zip](https://github.com/byte-physics/igor-unit-testing-framework/files/6290419/Notebook-Tracing-Chirp.zip)

Code:
```
Function Dostuff()

	string data, fName, procedure, text
	variable i, hit, err

	procedure = "MIES_AnalysisFunctions_PatchSeq.ipf"
	[data, fName] = LoadTextFile("c:trace:MIES_AnalysisFunctions_PatchSeq.ipf.log")

	WAVE/T matches = ListToTextWave(data, ";")
	Redimension/N=(-1, 2) matches
	
	matches[][1] = StringFromList(1, matches[p][0], ":")
	matches[][0] = StringFromList(0, matches[p][0], ":")

	[data, fName] = LoadTextFile("e:projekte:mies-igor:Packages:MIES:MIES_AnalysisFunctions_PatchSeq.ipf")

	data = ReplaceString("Z_();", data, "")
	
	// avoid wrapping and set text
	NewNotebook/F=1/N=Notebook0; err = GetRTError(1)
	Notebook Notebook0,ruler=Normal, textRGB=(0, 0, 0), margins={0,0,10000}, font="Anonymous Pro"
	Notebook Notebook0 selection={startOfFile, endOfFile}, ruler=Normal, text=data

	// mark everything uncovered
	// Notebook Notebook0 selection={startOfFile, endOfFile}, textRGB=(65535, 0, 0)

	// loop over notebook
	// and mark each line
	for(i = 0;; i += 1)
		FindValue/RMD=[][0]/TXOP=4/TEXT=(num2str(i)) matches
		if(V_Value == -1)
			text = "        | "
			Notebook Notebook0 selection={(i, 0), (i, 0)}, text=text //, textRGB=(65535, 0, 0)
		else
			sprintf text, "%.8#d| ", str2num(matches[V_Value][1])
			Notebook Notebook0 selection={(i, 0), (i, 0)}, text=text
			Notebook Notebook0 selection={(i, 0), (i + 1, 0)}, textRGB=(0, 65535, 0)
			hit += 1
		endif

		if(V_flag)
			break
		endif
	endfor

	printf "Coverage: %.1f %%\r", (hit/i) * 100
End
```

t-b:
@MichaelHuth I've added your original implementation of Z_() for playing around. But I don't think this is yet working. How should I use the optional `l` parameter?

MichaelHuth:
@t-b l is the optional parameter that is set when there was a line continuation and it counts the lines of the same code line.
When set then the current line to current line + l - 1 should be incremented in logging..

t-b:
@MichaelHuth Thanks.

MichaelHuth:
@t-b Also the function GetTracedProcedureNames() gets auto generated on instrumentation that allows to retrieve the mapping between procNum and procedure name. It returns a text wave, the index is the procNum.

t-b:
I've played around with coverage again. I've also adapted `Z_(...)`.

Some cases are IMHO a bit confusing:

```
00000003| 	if(ParamIsDefault(epochType))
00000001| 		return WB_GetStimsetParametersGlobal(setName)
        | 	else
00000002| 		return WB_GetStimsetParametersEpochType(setName, epochType)
        | 	endif
```

here both branches were executed but else and endif are never hit. Is that as intended?

MichaelHuth:
That is currently intended.

If that is confusing, we could add endif to the list of keywords where Z_ is added thereafter (like endswitch).
For the else, maybe the same, thus attributing it to the second execution block?

Internally there is no code executed for the "else" or "endif" line.

MichaelHuth:
@t-b Ready for some more testing.

It requires the TUFXOP for tracing and Igor 9.
I tested with RunAllTests.pxp from MIES and used in line 74 of UTF_Main.ipf
`RunTest(testsuite, name = name, enableJU = 1, debugMode= debugMode, enableRegExpTrace=1, traceWinList="MIES_.*\.ipf")`

For Igor 8 there seems to be a problem where Igor crashes in combination with the TUFXOP in the MIES tests.

It will output currently an rtf file for each traced procedure in the home path.

t-b:
@MichaelHuth Is it feasible to preserve the lineending in the instrumented files? In MIES all ipf files have LF but the instrumentation introduces CRLF. And people on the mac will use even other EOLs.

MichaelHuth:
At the moment I write 0x0D as line ending (CR).

Do you think it is important for the imho temporary instrumented files?
My requirement was that Igor is able to read it properly.

To keep the line endings I would have to track them per line and then probably duplicate it for all added lines.

t-b:
@MichaelHuth Well as it is currently, you can't really diff the files with git. I think you can assume that the EOLs are the same in the file. So if you take the one from the first line for the whole file this should be good.

t-b:
@MichaelHuth What is the left todo items for this PR?

I've come up with:
- [ ] Resolve crashes (2x), decide if we have to make it IP9 only
- [ ] Documentation: sphinx, Z_(), developer documentation for us
- [ ] Tests, I think having a short exemplararily file which is executed and we checked that the expected line numbers were hit
- [ ] Investigate cobertura support, see also https://github.com/byte-physics/igor-unit-testing-framework/issues/28#issue-382991468

t-b:
@MichaelHuth Running the MIES tests with IP8 and without tracing gives

```
Updated 7 paths from the index
Updated 0 paths from the index
unit-testing-tracing.ipf:29:35: error: Not yet available with ThreadSafe functions.

unit-testing-tracing.ipf:29:35: error: Not yet available with ThreadSafe functions.

unit-testing-tracing.ipf:29:35: error: Not yet available with ThreadSafe functions.

unit-testing-tracing.ipf:29:35: error: Not yet available with ThreadSafe functions.
```

so it looks like some ifdef igor version magic is missing.

t-b:
@MichaelHuth I'm still seeing in CI

```
unit-testing-tracing.ipf:29:35: error: Not yet available with ThreadSafe functions.
```

t-b:
And can we have something else than an abort if IP is too old. The current approch is not good for CI as a user has to click the acknowledgement button.

t-b:
Open issues:
- [ ] duplicated tag compilation error can occur quite easily
- [ ] Missing TFUXOP_Clear/N in Runtest. This is needed when calling it again during testing with instrumenting again.
- [ ] Make RTF creation optional, default to on when instrumenting

.gitlab-ci.yml
```diff
@@ -63,6 +63,72 @@ testing:
...
+    - exit $(grep -e "errors=\"[0-9]\+\"" -e "failures=\"[0-9]\+\"" tests/JU_*.xml | grep -cv "failures=\"0\" errors=\"0\"")
+  artifacts:
+    when: always
+    reports:
+      junit: tests/JU_*.xml
+    paths:
+      - tests/*.log
+      - tests/JU_*.xml
+      - screenshot/*.png
+
```
t-b, line 131: Can we integrate that in the testing job? Like in https://github.com/AllenInstitute/ZeroMQ-XOP/blob/4c5928d22928f51f30fdcc24e600c59592e7cd5b/.gitlab-ci.yml#L50

t-b:
This is already really working great! And will also hugely help in the future!!

Below are some comments which I wrote down while going through the code and the results in MIES.

Feel free to ping me for a chat once you went through the items and need some clarification.

Review:

- [ ] Various.pxp run sucessfully here. But if I call `run` the second time I get the dreaded TagFunc already
  exists error. Any solutions to this?
- [ ] RunTests: wnen -> when
- [ ] Can we invert the logic of `NORTFCREATION` and have `RTFCREATION` instead? Non-negated options are always easier to understand. `RTFCREATION` should default to true.
- [ ] The message "Tracing requires the Thread Utilities XOP" is not correct as we also need a recent version of IP9. The Build number for IP9 should also be a constant so that it can be referenced in that message and in all `#if` checks.
- [ ] What IP9 version are you using? Something beyond 37682? If yes we should also require that for tracing as this has
  a fix for the AfterCompiledHook reentrance issue.
- [ ] AfterCompiledHook: `if(!NVAR_Exists(ACHookEnabled) || (NVAR_Exists(ACHookEnabled) && !ACHookEnabled))` can be simplified by dropping the second `NVAR_Exists`
- [ ] WriteProcList has duplicated newlines
- [ ] AnalyzeTracingResult should handle TUFXOP_GetStorage erroring out gracefully
- [ ] AnalyzeTracingResult: `logdata += ..` Multithread?
- [ ] Documention CI job has `/builds/mirror/igor-unit-testing-framework/docu/sphinx/source/advanced.rst:407: WARNING:
  undefined label: traceoptionkeystrings`. This stems from the fact that the doxygen group `TraceOptionKeyStrings` does
  not have a rst label. I've hacked something into the docu building step in MIES, see
  AllenInstitute/MIES@ac30fb0cd689f93a499fd54c1416df63091463ed. It basically adds some code to the breathe generated rst
  files.
- [ ] Docu: The code for generating `rtfoutput.txt` should be present somewhere. Maybe in CoverageDemoCode.ipf? You should also add a comment near `..literalinclude:: rtfoutput.txt` how to regenerate it then.
- [ ]  Z_: `Each local thread storage contains of a 3d array` -> `Each local thread storage contains a 3d array`
- [ ] Can we set PROC_BACKUP_ENDING to `.backup` instead of `.traced`? After all this is a backup of the original file
  if I'm not mistaken.
- [ ] The RTF files should also have the `.ipf` suffix (which now is in the middle of the name) removed
- [ ] The instrumentation phase should already complain if MAX_PROC_LINES is too small
- [ ] Please move `Z_` into a separate file so that we can instrument the UTF in the future as well.
- [ ] Does it work on your end to view the rtf files in something else than IP? At least in Libreoffice/Wordpad I get
      a different textwidth which makes it impossible to read. I also played around with exporting into HTML via
      `SaveNotebook/S=5/H={"UTF-8", 0xFFFF, 0xFFFF, 0, 0, 32} Notebook2` and that works here. Any reason to prefer 
      RTF over HTML?
- [ ] I'm still not 100% happy with the if/else/endif coverage

```
00000102|00000000|00000102|		if(AI_SendToAmp(panelTitle, headStage, I_CLAMP_MODE, MCC_GETHOLDINGENABLE_FUNC, NaN))
________|________|________|			actualCurrent = AI_SendToAmp(panelTitle, headStage, I_CLAMP_MODE, MCC_GETHOLDING_FUNC, NaN, usePrefixes=0)
________|________|________|		else
00000102|________|________|			actualCurrent = 0
00000102|________|________|		endif
________|________|________|
00000102|________|________|		DEBUGPRINT("actualCurrent[A]=", var=actualCurrent)
________|________|________|
00000102|00000000|00000102|		if(!IsFinite(actualCurrent))
________|________|________|			print "Queried amplifier current is non-finite"
________|________|________|			ControlWindowToFront()
________|________|________|			continue
________|________|________|		endif
________|________|________|
00000102|________|________|		current += actualCurrent
```

The first else is not covered. And the second endif is also not covered. So in the first case the `else` belongs to the
true branch and endif to the false branch, and in the second case the endif belongs to the true case. This is probably
difficult to solve for the generic case, but how about taking into account that indentation is correct for finding the
corresponding else/endif/elseif statements?

- [ ] 
  
```
________|________|________|threadsafe static Function/WAVE HW_ITC_GetFifoPosFromConfig(config_t)
________|________|________|	WAVE config_t
________|________|________|
________|________|________|	Duplicate/FREE config_t, fifoPos_t
________|________|________|
00001260|________|________|	fifoPos_t[2,][] = NaN
00001260|________|________|	fifoPos_t[2][]  = -1
________|________|________|
00001260|________|________|	return fifoPos_t
________|________|________|End
```

is wrong as the Duplicate line must also be hit or?

Same with

```
________|________|________|Function ToggleCheckBoxes(win, checkBoxIn, checkBoxPartner, checkBoxInState)
________|________|________|	string win
________|________|________|	string checkBoxIn
________|________|________|	string checkBoxPartner
________|________|________|	variable checkBoxInState
________|________|________|
________|________|________|	SetCheckBoxState(win, checkBoxIn, checkBoxInState)
00000014|________|________|	DAG_Update(win, checkBoxIn, val = checkBoxInState)
00000014|________|________|	SetCheckBoxState(win, checkBoxPartner, !checkBoxInState)
00000014|________|________|	DAG_Update(win, checkBoxPartner, val = !checkBoxInState)
00000014|________|________|End
```

- [ ] Can we have the function start lines and End line always being covered when the function is executed?
- [ ] How easy is it to grey out unreachable code in the RTF output? Like code outside functions. This would
  help to guide the eye when reading.
- [ ] Empty lines are currently never covered. Did we decide that last time we talked about that?
- [ ] CoverageDemocode.ipf should probably live in docu/examples or?

t-b (changes requested):
See comment.

MichaelHuth:
For else, Z_ after.

For endif Z_ after.

t-b:
The reason the sphinx documentation warning did not show up as CI error was that I messed up the logic. This is now fixed in #208. So if you rebase against main you get some CI errors ;)

t-b:
Is it currently supported to instrument test suites? If yes is the tag function treated as test case? Maybe that needs a `_IGNORE` suffix then?

MichaelHuth:
> Is it currently supported to instrument test suites? If yes is the tag function treated as test case? Maybe that needs a `_IGNORE` suffix then?

The only limitation for instrumentation is currently that all procedure files starting with "unit-testing" are excluded. (For tracing the same filter as for RunTest is used). I can add the _IGNORE suffix to prevent clashes with test suites.

t-b:
>  I can add the _IGNORE suffix to prevent clashes with test suites.

+1

t-b:
Here is the list of changes we talked about earlier this week:
- [ ] Remove/Rename variables mentioning RTF instead of HTML
- [ ] Fix indentation in HTML coverage output
- [ ] Use operation queue for starting RunTest from AfterCompileHook so that other AfterCompileHooks can run as well
- [ ] Get HTML as artefact in CI tests
- [ ] ... (Things I forgot?)

Issue #211, opened 2021-09-02
UTF generates invalid JUNIT xml file


t-b:
Found at https://github.com/AllenInstitute/MIES/pull/1045, and there the "Test from repo" test has

```
Failed to parse test result file "C:\Users\svc_multipatch\bamboo-agent-home\xml-data\build-dir\MIES-PLAYIN466-CFR\tools\unit-testing\JU_RunAllTests_2021-09-02_05-03-01.xml"
```

Source: http://bamboo.corp.alleninstitute.org/browse/MIES-PLAYIN466-CFR-1 (requires VPN)

I'm attaching the file, see [JU_RunAllTests_2021-09-02_05-03-01.zip](https://github.com/byte-physics/igor-unit-testing-framework/files/7101902/JU_RunAllTests_2021-09-02_05-03-01.zip).

Running it through xmllint gives

```
E:/JU_RunAllTests_2021-09-02_05-03-01.xml:254: parser error : Opening and ending tag mismatch: testcase line 253 and failure
" type="FAIL"></failure>
                        ^
E:/JU_RunAllTests_2021-09-02_05-03-01.xml:257: parser error : Opening and ending tag mismatch: testsuite line 253 and testcase
                </testcase>
                           ^
E:/JU_RunAllTests_2021-09-02_05-03-01.xml:259: parser error : Opening and ending tag mismatch: testcase line 253 and failure
" type="FAIL"></failure>
                        ^
E:/JU_RunAllTests_2021-09-02_05-03-01.xml:262: parser error : Opening and ending tag mismatch: testsuites line 253 and testcase
                </testcase>
                           ^
E:/JU_RunAllTests_2021-09-02_05-03-01.xml:263: parser error : Extra content at the end of the document
                <testcase name="ConfigurationTest#TCONF_AllStates in UTF_Configuration.ipf (3)
                ^
```

and the problem is that e.g.

```xml
		<testcase name="ConfigurationTest#TCONF_RelevantValues in UTF_Configuration.ipf (3)" classname="ConfigurationTest#TCONF_RelevantValues" time="0.347">
" type="FAIL"></failure>
```

is not valid xml.

t-b:
This can very well be a CI/bamboo issue.

MichaelHuth:
I think it is some kind of Igor issue, the code line generating it is not different from the one for an error,
so
[...]

PR #212, opened 2021-09-03
fix for cutoff output in xml file for messages > 2400 characters

MichaelHuth:
- sprintf has in IP8 a limit of 2400 characters. There were use cases where
  the failure message had more characters per line (e.g. wave compare)
- This broke the CI

close https://github.com/byte-physics/igor-unit-testing-framework/issues/211


MichaelHuth:
@t-b I think this should fix these cases and still allows to diff the output from long wave compares.

t-b:
@MichaelHuth Thx. This should be it.

t-b (approved):
LGTM

PR #222, opened 2021-11-21
Add GetWavePointer

t-b:
This is a generic function to return the memory address of a wave which
works on all Igor Pro versions.


MichaelHuth (approved):
Thats a good trick.

PR #226, opened 2021-12-22
RunTest: Allow passing an empty traceWinList

t-b:
We currently always do instrumentation as soon as the optional
traceWinList parameter is set.

But that is not nice for generic code calling us as that means it has to
branch out.

Accept an empty traceWinList string meaning the same as it being not
present.


MichaelHuth (changes requested):
Thats good, though the docu needs also a correction then in advanced.rst somewhere near line 393.

t-b:
@MichaelHuth Thx. Changec.

MichaelHuth (approved):

Issue #198, opened 2021-06-07
Verify if DebugState set Debugger properly when tests run.


MichaelHuth:
We had in some tests the problems that the Debugger was off, where it should have been enabled.

Proposed change:
We map allowdebug completely to debugMode and remove allowDebug from the structure.
debugMode represents the debugger state while the IUTF runs.
allowDebug means that the debugger state at call time is saved in debugMode.
Remove also IUTF_DEBUG_ALLOW constant then as debugMode needs no separate flag for allowDebug then.
debugMode = 0 means that the Debugger is off during IUTF run.

PR #227, opened 2022-01-13
Code Cleanup for handling of allowDebug parameter

MichaelHuth:
allowDebug is now handled through debugMode internally.

debugMode sets the state of the debugger while a testcase runs,
if debugMode is set.

Thus, allowDebug just means that the debugger state while the IUTF is running
is the same as outside the IUTF. This is set now accordingly.

Also added Tests to track the Debugger Settings through a test run.

close https://github.com/byte-physics/igor-unit-testing-framework/issues/198


t-b:
Nice!

t-b (approved):

PR #228, opened 2022-01-24
Add line numbers in front of tracing result html output

MichaelHuth:
Adds line numbers in front of the output in the html files from tracing.


t-b:
Nice!

t-b (approved):

Issue #207, opened 2021-07-14
Output total coverage for gitlab to stdout


t-b:
See https://docs.gitlab.com/ee/ci/pipelines/settings.html#test-coverage-parsing.

PR #230, opened 2022-01-24
Add coverage output to history and stdout at end of traced run

MichaelHuth:
- The statistics is internally sampled by procedure file, a global result is
  output

- Functions marked with the UTF_NOINSTRUMENTATION tag are included in the
  code lines number.

close https://github.com/byte-physics/igor-unit-testing-framework/issues/207

### DIFF
--- a/docu/sphinx/source/advanced.rst
+++ b/docu/sphinx/source/advanced.rst
@@ -436,6 +436,10 @@ The evaluation of gathered coverage data refers to the procedure file content on
 in procedure files that are targeted for instrumentation will result in incorrect result files. It is strongly recommended to save all
 procedure file changes to disk before running a test with code coverage logging.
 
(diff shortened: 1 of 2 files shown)


t-b (changes requested):
See comment.

MichaelHuth:
Remark to self: The p-value for the percentage includes all code lines.

Line counting should be correct now.

t-b:
I've updated the branch from the linked PR above. Looking at the HTML output,
[MIES_ThreadsafeDataSharing.zip](https://github.com/byte-physics/igor-unit-testing-framework/files/7933498/MIES_ThreadsafeDataSharing.zip) I would expect that to have 100% line coverage, but I only get `75.9%`. I suspect the empty lines being the culprit.

MichaelHuth:
The empty lines are now excluded from counting for coverage (they are still black in the html)

t-b (approved):
Nice!

Issue #214, opened 2021-10-18
Tracing: elseif not handled properly


t-b:
```
00000920|00000833|00000087|            if(dataAcqOrTP == DATA_ACQUISITION_MODE)

00000833|________|________|                numDACs = DC_NoOfChannelsSelected(panelTitle, CHANNEL_TYPE_DAC)

00000833|________|________|                numADCs = DC_NoOfChannelsSelected(panelTitle, CHANNEL_TYPE_ADC)

00000833|________|________|                numTTLsRackZero = DC_AreTTLsInRackChecked(panelTitle, RACK_ZERO)

00000833|________|________|                numTTLsRackOne = DC_AreTTLsInRackChecked(panelTitle, RACK_ONE)

00000087|________|________|            elseif(dataAcqOrTP == TEST_PULSE_MODE)

00000087|________|________|                numActiveHeadstages = DC_NoOfChannelsSelected(panelTitle, CHANNEL_TYPE_HEADSTAGE)

00000087|________|________|                numDACs = numActiveHeadstages

00000087|________|________|                numADCs = numActiveHeadstages

00000087|________|________|                numTTLsRackZero = 0

00000087|________|________|                numTTLsRackOne = 0

________|________|________|            else

________|________|________|                ASSERT(0, "Unknown value of dataAcqOrTP")

00000920|________|________|            endif
```

I'm not understanding why the elseif does not have any entries in the second/third coverage column. It has in the documentation.

PR #231, opened 2022-01-25
bugfix: lines starting with elseif got instrumented as lines with else

MichaelHuth:
Since commit de9e739cd0492ddd2432c01a53a658767494197e

Lines starting with elseif were instrumented the same as lines with else,
where a Z function was added in the following line. The condition of the elseif
line was not replaced.
The cause was that the mentioned commit changed the handling of lines starting
with else to the described behavior. However, lines with keys, such as else,
were checked before conditions like elseif/if. Since, "elseif" also starts with
"else" it was misidentified as "else".

The solution is to change the priority order for adding the Z_ function.
Now conditions from lineStartZReplaceKeys are checked first.

close https://github.com/byte-physics/igor-unit-testing-framework/issues/214


t-b:
- [ ] Missing tests

t-b (changes requested):
See comment.

t-b (approved):

Issue #238, opened 2022-03-01
Created JUNIT XML files don't validate against our xsd schema


t-b:
```
$ xmllint --version
xmllint: using libxml version 20910
   compiled with: Threads Tree Output Push Reader Patterns Writer SAXv1 FTP HTTP DTDValid HTML Legacy C14N Catalog XPath XPointer XInclude Iconv ICU ISO8859X Unicode Regexps Automata Schemas Schematron Modules Debug Zlib Lzma 
```

No basic XML errors:

```
$ xmllint --noout junit.xsd JU_HardwareTests-ITC1600_2022-02-08_15-32-33.xml
```

This is xmllint from debian bullseye, the one from git for windows does not finish (>= 2h).

```
$ xmllint --noout --schema junit.xsd JU_HardwareTests-ITC1600_2022-02-08_15-32-33.xml
JU_HardwareTests-ITC1600_2022-02-08_15-32-33.xml:3: element testsuite: Schemas validity error : Element 'testsuite', attribute 'disabled': The attribute 'disabled' is not allowed.
JU_HardwareTests-ITC1600_2022-02-08_15-32-33.xml:25: element testsuite: Schemas validity error : Element 'testsuite', attribute 'disabled': The attribute 'disabled' is not allowed.
JU_HardwareTests-ITC1600_2022-02-08_15-32-33.xml:38: element testsuite: Schemas validity error : Element 'testsuite', attribute 'disabled': The attribute 'disabled' is not allowed.
JU_HardwareTests-ITC1600_2022-02-08_15-32-33.xml:108: element testsuite: Schemas validity error : Element 'testsuite', attribute 'disabled': The attribute 'disabled' is not allowed.
JU_HardwareTests-ITC1600_2022-02-08_15-32-33.xml:130: element testsuite: Schemas validity error : Element 'testsuite', attribute 'disabled': The attribute 'disabled' is not allowed.
JU_HardwareTests-ITC1600_2022-02-08_15-32-33.xml:567: element testsuite: Schemas validity error : Element 'testsuite', attribute 'disabled': The attribute 'disabled' is not allowed.
JU_HardwareTests-ITC1600_2022-02-08_15-32-33.xml:586: element testsuite: Schemas validity error : Element 'testsuite', attribute 'disabled': The attribute 'disabled' is not allowed.
[...]

PR #240, opened 2022-03-02
Remove optional JUNIT TestSuite attribute "disabled"

MichaelHuth:
The strict junit.xsd definition does not know the optional attribute
"disabled". Also, we did not use it in the IUTF and it was always zero.

To allow verification against the xsd xml definition file, the attribute is
removed with these changes.

close https://github.com/byte-physics/igor-unit-testing-framework/issues/238

@t-b Please check with the Debian bullseye xmllint as my xmllint version here has the infinite runtime issue.


t-b:
> Please check with the Debian bullseye xmllint as my xmllint version here has the infinite runtime issue.

@MichaelHuth This should be solved with CI as already mentioned in https://github.com/byte-physics/igor-unit-testing-framework/issues/238#issuecomment-1056984821. For CI just grab a debian docker image and install xmllint.

t-b (approved):

Issue #234, opened 2022-01-31
Tracing and macros


t-b:
In MIES we have a couple of panel macros. And it looks like macros are treated as uncovered as in 
[MIES_DebugPanel.zip](https://github.com/byte-physics/igor-unit-testing-framework/files/7971833/MIES_DebugPanel.zip).

Check if it is possible to also add trace statements inside macros.

PR #242, opened 2022-03-05
Support macros in instrumentation

MichaelHuth:
Changes to support also Macros in instrumentation.

close https://github.com/byte-physics/igor-unit-testing-framework/issues/234


MichaelHuth:
@t-b I could use some information from the CI machine. Currently the tests on Various.pxp time out. If the job times out no screenshot is generated. The Various.pxp test works here locally.

t-b:
@MichaelHuth 
grafik

MichaelHuth:
@t-b ok, the cause was that this line already relies on the auto-indexing skip on zero size patch in the recent nightlies of IP9. Thus, with the nightly this worked out of the box.

t-b (approved):
Very nice!

t-b (approved):

PR #250, opened 2022-05-06
Fix: Tracing must parse function decl types case-insensitive

MichaelHuth:
Tracing parsed the function declaration types case-sensitive
Thus, for declaration if "variable" declaration is written as "Variable"
then the attribution of lines to the declaration block is aborted early.

since de9e739cd0492ddd2432c01a53a658767494197e


t-b (approved):
Nice find!

Issue #254, opened 2022-07-02
CHECK_EQUAL_WAVES ignores mode for zero sized waves


MichaelHuth:
Example:

```
Function Test()
  Make/FREE/N=0 w1
  Make/FREE/D/N=0 w2
  CHECK_EQUAL_WAVES(w1, w2, mode = WAVE_DATA | DIMENSION_SIZES)
End

```

Assertion fails because the wave type is different, but the wave type should not be considered according to the mode setting.

t-b:
We've reported that to WM already. And this is marked as wontfix. I'd be happy to add a workaround for this exact case.

PR #255, opened 2022-07-04
Compare zero sized waves properly with set mode in EQUAL_WAVES_WRAPPER

MichaelHuth:
The Igor Pro function "equalwaves" does return a RTE when the wave data should
be compared and one wave is of zero size.
For that we caught that condition beforehand and did a simple wave type check
independently of the mode set.

This was crude and let to confusion when actually using mode settings.

This commit fixes that behavior with two changes:
- The check for compatible wave types for Igor Pros equalwaves was adjusted
  to be applied only if at least one wave has data points.
  Igor Pros EqualWaves throws an RTE when Textual and Numerical waves should be
  compared.
- Otherwise UTF_Checks#AreWavesEqual is called with each single mode setting
  separately also for zero sized waves. In that function we handle the case
  mode == WAVE_DATA manually and do no call Igor Pros EqualWaves.

close https://github.com/byte-physics/igor-unit-testing-framework/issues/254


t-b:
Please add some tests to VTEE.ipf as well, it has an `AreWavesEqual` section already. Thanks.

MichaelHuth:
Updated

t-b:
@MichaelHuth Nice! Thanks.

t-b (approved):

Issue #168, opened 2021-02-09
UTF_TD_GENERATOR functions are called way too often


t-b:
```igorpro
#include "unit-testing"

Function/WAVE Gen()

	print "called"
	
	Make/FREE/N=1 data
	return data
End

// UTF_TD_GENERATOR Gen
Function Test([variable var])

	PASS()
End
```

```
•runTest("procedure")
  called
  called
  Start of test "Unnamed"
  called
  called
  Entering test suite "Procedure"
  called
  Entering test case "Test:0"
  called
  Leaving test case "Test:0"
  Finished with no errors
  Leaving test suite "Procedure"
  Test finished with no errors
  End of test "Unnamed"
```

Change that, so that it is only called once and add CI tests to verify and keep it like that.

MichaelHuth:
In getTestCasesMatch the function CheckFunctionSignaturesTC is called (which calls the DataGenerator) and afterwards GetTestCaseCount (which also calls the DataGenerator). These information could be gathered combined, by e.g. returning two lists.

Also in CheckFunctionSignaturesTC the fullTestCase can be added to the output list, such that building the list in getTestCasesMatch can be removed.

MichaelHuth:
Latest approach idea:

In the beginning build text wave in getTestCasesMatchwith all relevant information for test cases.

Example fields:
procWin0 testcase0 fullFunctionName list_of_datagenerators list_of_dgen_wavesizes skip? expected_fail
procWin0 testcase1 fullFunctionName list_of_datagenerators list_of_dgen_wavesizes

Additionally build data generator name to returned wave mappig wave (wave ref wave where dimLabels are dgen Names)
dgenName:WaveRef -> returned wave

Determine if function sig fits to multimultidata testcase ,in begin of CallTestCase for this specific funcsig, add if branch

In main loop instead of running over procWin and funcNames, run only over index of above text wave that contains one row per testcase.

Goal:
Parsing of function header only once.
Calling of multidata generators only once.

Current concept start:
Add creation of above text wave in current code transparently.
then
Replace locations where information is currently parsed/gathered to look it up in our new wave.

PR #262, opened 2022-09-02
Revise IUTF setup step

MichaelHuth:
Optimize code by setting up global waves that store most of the data per test case.
Also add global wave to store datagenerator waves and tag values of functions.

Change code to use these new global storages and discard the old obsolete ones.

close https://github.com/byte-physics/igor-unit-testing-framework/issues/168

### DIFF
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -1814,8 +1814,11 @@ End
 /// Function determines the total number of test cases
 /// Normal test cases are counted with 1
 /// MD test cases are counted by multiplying all data generator wave sizes
+/// When the optional string procWin is given then the number of test cases for that
+/// procedure window (test suite) is returned.
 /// Returns the total number of all test cases to be called
-static Function GetTestCaseCount()
+static Function GetTestCaseCount([procWin])
+	string procWin
 
 	variable i, j, size, dgenSize
 	variable tcCount, dgenCount
@@ -1825,6 +1828,10 @@ static Function GetTestCaseCount()
 	WAVE/T testRunData = GetTestRunData()
 	size = DimSize(testRunData, UTF_ROW)
 	for(i = 0; i < size; i += 1)
+		if(!ParamIsDefault(procWin) && CmpStr(procWin, testRunData[i][%PROCWIN]))
+			continue
+		endif
+
 		dgenCount = 1
 		dgenList = testRunData[i][%DGENLIST]
 		dgenSize = ItemsInList(dgenList)
@@ -3052,8 +3059,7 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
 			getLocalHooks(s.procHooks, procWin)
 
 			if(startNextTS)
-				// TODO need tcCount for procWin
-				s.juProps.testCaseCount = s.tcCount
+				s.juProps.testCaseCount = GetTestCaseCount(procWin=procWin)
 				ExecuteHooks(TEST_SUITE_BEGIN_CONST, s.procHooks, s.juProps, procWin, procWin)
 				s.juProps.testSuiteNumber += 1
 			endif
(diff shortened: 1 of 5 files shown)


t-b:
Review:

57064d76 (Add wave getter for testrun data, data generator and utility function, 2022-09-02)

```
// Simple function to automatically increase the wave size by a chunk
//
EnsureLargeEnoughWaveSimple
```

The documemntation should say that we increase the number of rows only.

e39567e8 (Fill test run data wave, 2022-09-02)

```
	EnsureLargeEnoughWaveSimple(testRunData, tdIndex + 1)
```

In MIES we always call EnsureLargeEnoughXXX with just tdIndex as the function ensure that this is a valid index into the wave. I think this would work here as well or? 

main has

```
				if(TAP_IsOutputEnabled())
					s.tap_skipCase = TAP_IsFunctionSkip(s.fullFuncName)
				endif
```

but here we have 

```
				testRunData[tdIndex][%TAP_SKIP] = num2istr(TAP_IsFunctionSkip(fullFuncName))
```

I think the SKIP tag should only be evaluted if TAP output is enabled.

89a886de (The testcase count is now determined once through the test setup data, 2022-09-06)

I think you can drop the special handling of an empty dgenList in GetTestCaseCount as the loop below works as well in this case.

And in which case is

```
+                       dPos = FindDimLabel(dgenWaves, UTF_ROW, dgen)
+                       if(dPos == -2)
+                               print "Data Generator not found in wave."
+                               Abort
+                       endif
+
```

triggered?

Please also add function level documentation for GetTestCaseCount().

e2bc8f03 (Use global wave to store function tags and their values, 2022-09-06)

Is there a reason you call AddFunctionTagWave before CheckFunctionSignatureTC
in CreateTestRunSetup? I would expect it to be called after
CheckFunctionSignatureTC as we only want to fill it with tags for valid test
cases.

c40b82a6 (Adapted main loop to index over test run data wave instead of two loops, 2022-09-06)

Typo in commit message:  "folde rin"

I think

```
	if(EnabledDebug())
		EnableDebugOutput()
	endif
```

can be dropped as it does nothing now. 

This needs the previousProcWin fix we talked about in the other PR. And also the missing

```
               if(shouldDoAbort())
                       break
               endif

```

after `while(...)`.

The comment

```
		// Kills data folder and reinitializes
		ExecuteHooks(TEST_BEGIN_CONST, s.hooks, s.juProps, s.name, NO_SOURCE_PROCEDURE, param=s.debugMode)
```

is wrong as you removed the folder removal.


```
		if(!reentry && CmpStr(s.previousProcWin, procWin))

    [...]

			s.procHooks = s.hooks
			// 3.) get local user hooks which reside in the same Module as the requested procedure
			getLocalHooks(s.procHooks, procWin)
```

In main we do call getLocalHooks always when !reentry is true.

The two reentry blocks at the top of the 

 ```
	for(i = 0; i < tcFuncCount; i += 1)
 ```

can then also be joined.

Can we remove `s.tap_caseCount` already? 

726f757c (Retrieve Data Generator for TC from Setup Wave, 2022-09-06)

Nice.

2d2dc50c (Add local tap_skipcase variable with data from TestRunSetup wave, 2022-09-06)

Nice!

131b7c8c (Get correct testcase count per testsuite for JUNIT data, 2022-09-06)

Function level documentation for GetTestCaseCount.

02d6c3bf (Remove requirement of allTestCases in main loop returned by getTestCasesMatch, 2022-09-06)

```
+                       sprintf msg, "Error %d in getTestCasesMatch: %s", err, str
```

The function getTestCasesMatch does not exist anymore.

Can we call str in

```
		err = CreateTestRunSetup(s.procWinList, s.testCase, s.enableRegExpTC, str)
```

something like errMsg?

851d2f98 (Remove unused entries of strRunTest structure, 2022-09-06)

I think testcaselist and tcCount can also go.

4fb29772 (Adapt tracing logic to get also tags for traced functions, 2022-09-07)

I think this needs to be part of e2bc8f03 (Use global wave to store function tags and their values, 2022-09-06) as it fixes this commit or?

51e58687 (Set expected failure flag from TestRunData wave, 2022-09-07)

Good!

ce55723f (Adapt VTTE tests to changed API of former getTestCasesMatch function, 2022-09-06)

Jeep!

I would also like to see a new test added to Various.ipf which checks that the generator function is only called once.

MichaelHuth:
@t-b 
> Is there a reason you call AddFunctionTagWave before CheckFunctionSignatureTC
> in CreateTestRunSetup? I would expect it to be called after
> CheckFunctionSignatureTC as we only want to fill it with tags for valid test
> cases.

The test case checked in CheckFunctionSignatureTC could be a multi data test case where GetDataGenFullFunctionName is called to determine the data generator function name, there the tags of the test case get evaluated. So the tags have to be added before.

MichaelHuth:
Ready for next review.

MichaelHuth:
Good catch. I fixed that.

t-b (approved):

Issue #213, opened 2021-10-18
Implement ways of having multiple generator functions per test case


t-b:
The easiest option:
- Allow multiple optional parameters of different types
- Require the generator functions be listed in the order of the optional parameters
- Iterate through all possibilities when calling the testcase

```igorpro

// UTF_TD_GENERATOR GetStrings, GetVariable
static Function MSQ_DS1([string str, variable opt])

// do something

End

```

t-b:
So I went the route you suggested in https://github.com/AllenInstitute/MIES/pull/1093/commits/b0393e6619ce6a12b30962228f7a1138eb817389.

This is a MIES hardware test were I want to execute the test for all MD devices and various double values.

Alone the generator function is ugly to write, and the parameter extraction is also always a few lines of code. This does also not support the case were `devices` has multiple entries.

If you're main concern is 

> we need a lot of new reserved function signatures.

I'm open to suggestions how to cut the number of reserverd function signatures down. Maybe allow structs as parameters?

t-b:
Thanks for the writeup!

> The specification string contains a double dot separating 

When thinking about it now, I think I would prefer separating that with a space instead of a double dot/colon. Do you think this would cause any problems?

> The test case will be named testCase:<arg0>:<arg1>:<arg2> representing the values of the active arguments separated by :.

This might be implied here, but I wanted to mention that this should also support dimension labels of the generator waves like we suppport now.

From the writeup it looks like we would call the generator functions once for each test case execution, see also #168. Is that correct? Should we fix that first?

t-b:
> The IUTF will check if there is a // UTF_TD_GENERATOR key specified within the first 4 lines before the test case function declaration and will increase the "look before" on demand to gather all specifications.

I think we should just change that to look into the complete function comment. Would that work?

t-b:
Can you add a time estimate as well?

Issue #158, opened 2020-10-10
Data generator function returning empty waves should be reported as skipped in TAP/JUNIT


t-b:
Successor of #99 and https://github.com/byte-physics/igor-unit-testing-framework/pull/131.

```igor

function/WAVE generator()

	make/free/N=0/T data

	return data
end


/// UTF_TD_GENERATOR generator
Function MyTest([str])
	string str

End

Function dostuff()

PASS()

End
```

This should do the following:
- no warning about zero size points
- message to history that the test case is skipped
- Mark as skipped in JUNIT/TAP

This probably requires to generalize strRunTest.tap_skipcase as we now also have skipped test cases for non-TAP usage. It might also make sense to let ExecuteHooks handle skipped test cases which makes it call sites more readable. We would have to call `JU_TestCaseBegin` and `JU_TestCaseEnd` for skipped cases as well.

PR #263, opened 2022-09-08
Add new multi multi data test cases with multiple input variables

MichaelHuth:
Adds Multi-Multi data test cases

close https://github.com/byte-physics/igor-unit-testing-framework/issues/213
close https://github.com/byte-physics/igor-unit-testing-framework/issues/158


t-b:
Review: 
da3fa091 (Add structure and test case prototype function for multi-multi-data tests, 2022-09-08)

Jeep.

f9c54c45 (Adapt tagWave creation to store mmData generator tags with unique dimLabel, 2022-09-08)

Hmm.

77bbfc7f (Add data generator function validity check for mmData Generators, 2022-09-09)

```
+    FUNCREF TEST_CASE_PROTO_MD fTCmmd = $fullFuncName
+    if(UTF_FuncRefIsAssigned(FuncRefInfo(fTCmmd)))
+        dgenList = CheckFunctionSignatureMDgen(procWin, fullFuncName)
+        return 0
+    endif
+
```

Newline before this block.

Note to self: CheckFunctionSignatureTC now aborts with an empty generator wave.

62021e62 (Added constants for mode for multi data test cases, 2022-09-09)

Nice find.

9cbaa389 (Add MMD TestCase setup and Call, 2022-09-09)

Does GetMMDataState store GetMMDFuncState waves? If yes the documentation should say so.

ea31c6d9 (Add test case suffix for MMD test cases, 2022-09-09)

```
+        if(index < genSize)
+            mdFunState[i][%INDEX] = num2istr(index)
+            break
+        else
+            mdFunState[i][%INDEX] = num2istr(0)
+        endif
+    endfor
+
+    return i == numVars
+
```

That is difficult to follow. How about

```
for(...)
  [...]
  if(index < genSize)
    mdFunState[i][%INDEX] = num2istr(index)
    return 0
  endif

  mdFunState[i][%INDEX] = num2istr(0)
endfor

  return 1
End
```

instead?

58d91183 (Introduce StrConstant for test case suffix separator, 2022-09-09)

Okay.

5a181942 (Change GetFunctionTagWave to allow upto numLines tags, 2022-09-09)

Nice.

5ffaab41 (Call data generator only once for MMD test cases, 2022-09-09)

Yes!

4e78d987 (Check MMDataGen tags for correct variable names, 2022-09-12)

```
+                testRunData[tdIndex][%TAP_SKIP] = num2istr(TAP_IsFunctionSkip(fullFuncName) | markSkip)
```

Logical or instead for clarity?

c0ee0ce9 (Mark TC with zero sized data generator for SKIP, 2022-09-12)

Juup.

8e9bd0e9 (Rename Dimlabel of TestRunData from TAP_SKIP to SKIP, 2022-09-12)


Unrelated change:

```
-            ExecuteHooks(TEST_SUITE_BEGIN_CONST, s.procHooks, s.juProps, procWin, procWin)
+            ExecuteHooks(TEST_SUITE_BEGIN_CONST, s.procHooks, s.juProps, procWin, procWin, s.i)
```

6e21fedc (Count test suite numbers correctly for JUNIT output, 2022-09-12)

Nice find!

3b11806a (Evaluate SKIP in junit as skipped test case, 2022-09-12)

Good!

b9146587 (Replace tap_caseCount with loop index, move into ExecuteHooks function, 2022-09-12)

+1

35c81385 (Determine previousProcwin locally from previous TestRunData index, 2022-09-12)

Should be part of #262 and squashed into the commit introducing previousProcWin.

6e230e8b (Prefer s.i over i in index variable usage in main loop, 2022-09-12)

Okay.

7bda0d56 (Fix: Abort check after MD TC loop was missing, 2022-09-12)

Should be part of #262 and squashed into the commit introducing the bug.

14886829 (Rename tap_skipCase to skip, 2022-09-12)

Should be part of 3b11806a.

9b886b23 (Fix: get wGenerator reference correctly after reentry for tcSuffix, 2022-09-12)

Please squash that into the commit introducing the bug.

7815869b (Add tests for Multi-Multi-Data Test Cases, 2022-09-09)

Can you add these tests to Various.ipf? I would like to reserve VTTE.ipf for test assertion checks.
And can you also verify that the data generator functions are only called once?

3e6a37dd (Add documentation for new feature and changed behavior, 2022-09-12)

```
+If the TAP directive to skip a test case was placed then the test case is also
+treated to be skipped when JUNIT is enabled.
```

I think this not desired but can be presuaded otherwise.


```
+       int64 i3
+      int64 i4
+   EndStructure
+
```

Wrong indentation.


```
+If Multi-Multi-Data test cases are combined with functions with background activity the reentry function must have the same
+declaration.
+
```

declaration -> signature.


```
change fastest
```

-> changes

t-b (changes requested):
See comment.

t-b:
I think this PR also closes #158.

MichaelHuth:
> Logical or instead for clarity?

I think a boolean OR is correct here because effectively TAP_IsFunctionSkip() returns a boolean and markSkip is also a boolean and both boolean flags are of the same type "skip".
Should we have some recommendation in the coding convention when to prefer logical ORs?

t-b (approved):

t-b:
LGMT.

PR #267, opened 2022-10-04
RunTest: Initialize at the earliest possible moment

t-b:
When running testing with instrumentation and the instrumented code errors out as the code is already instrumented we don't output any history log. This happens with all errors which happen earlier as the history log initialization.

So let's move the initialization to the earliest possible moment.


t-b:
@MichaelHuth Please have a look.

PR #268, opened 2022-10-07
Fix #266. Test case hooks are evaluated as MD test cases in Igor Pro 6

### LINKED ISSUE #266: Test case hooks are evaluated as MD test cases in Igor Pro 6
A test case hook like:

```
static Function TEST_SUITE_END_OVERRIDE(name)
	string name

End

```

created a valid FUNCREF with the MD test case prototype function
```
Function TEST_CASE_PROTO_MD_STR([str])
	string str

```

in Igor Pro 6. Thus, the hook is evaluated as potential MD test case.

Usually the test run is aborted because no data generator was specified for it.

Garados007:
Every function is now checked to have only optional parameter. This will prevent hooks from used as testcases.

Close #266 .

### DIFF
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -1629,9 +1629,16 @@ static Function CheckFunctionSignatureTC(procWin, fullFuncName, dgenList, markSk
 
 	variable err, wType1, wType0, wRefSubType
 	string dgen, msg
+	string funcInfo
 
 	dgenList = ""
 	markSkip = 0
+
+	// Require only optional parameter
+	funcInfo = FunctionInfo(fullFuncName)
+	if (NumberByKey("N_PARAMS", funcInfo) != NumberByKey("N_OPT_PARAMS", funcInfo))
+		return 1
+	endif
 
 	// Simple Test Cases
 	FUNCREF TEST_CASE_PROTO fTC = $fullFuncName
--- a/tests/Various.ipf
+++ b/tests/Various.ipf
@@ -346,11 +346,9 @@ Function TC_MMD_bck_REENTRY([md])
 	PASS()
 End
 
-#if IgorVersion() >= 7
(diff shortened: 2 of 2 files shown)


t-b:
Review:

033ec55c (Fix #266., 2022-10-07)

- The change itself looks good, requiring that all parameters to test cases are optional is the right thing to do.
- The commit message is a bit too terse. First of all the subject line should mention the code area/subsystem the change is touching and summarizing the change. Mentioning the issue in the subject is not good, as it requires the reader to jump to another place, this can be done in the body at the end. The rest of the commit message should explain *why* the change is good, so basically the motivation. See also https://cbea.ms/git-commit.

Some good examples here in the repo are: 9b2dfd53 (Documentation: Use matching case for group names, 2022-07-06), 804bc38c (CI: Add html from tests in artefacts., 2021-07-23) and 361d761d (add regexp functionality to RunTest, 2018-04-08)

- We also favour explaining things in commit messages vs. comments in the code. The reason is that it does not clutter the code but still allows to retrieve the info. In this case here I would change the comment to "All test cases must only have optional parameters" and explain the rest in the commit message. Note also "must" vs "should".
- We can easily add a test case here as having an test hook without test assertions should already trigger an error with IP6. Looking at 

https://github.com/byte-physics/igor-unit-testing-framework/blob/caadcd6cd4bbf83d6a3a423efce9d12221f00152/tests/Various.ipf#L349-L356

even shows that this test code works around this exact bug you are fixing here.
- Tests usually should be part of the commit changing the code, as that enables bisectability of the repository. But that is a real should as we are not requiring it.

t-b (changes requested):
See comment.

Garados007:
I have updated the git commit message, changed the comment and updated the test case.

t-b:
Nice. LGTM.

t-b (approved):

PR #271, opened 2022-10-10
Documentation: Fix the number of tags

### LINKED ISSUE #269: Fixing max. 4 tags for a function in the documentation
In the documentation are several parts like:

```md
(maximum 4 lines above `Function` are considered as tags, every tag in separate line)
```

This limit was removed as the MMD test cases are introduced.

Garados007:
There were a few places in the documentation where it said that a maximum of 4 tags per function are recognised.

This is deprecated since the introduction of MMD test cases. Now all lines up to the previous function can have tags.

This change reflect this new behavior in the documentation.

Close #269.


MichaelHuth (approved):
Looks good.

Issue #221, opened 2021-11-19
Enhance difference printing for equal waves


t-b:
- Output the name for free waves if present. Since IP9 free waves can have names via
```
Function Dostuff()
	Make/FREE=1 important_name
	print NameOfWave(important_name)
End
```
- Drop unused trailing `[0]` dimensions in difference table
- Comparing the output of text wave with differing strings can be hard. Maybe we can provide a colorized diff of the string as well?
- Can we remove trailing zeros for integers in
```
Dimensions  |Labels|Value    |
------------|------|---------|
[3][0][0][0]|      |15.000000|
            |      |17.000000|
```
- Can we also display enough digits so that the there is always a difference visible?

t-b:
Last point: print number to string with increasing digit number until diff is present

Garados007:
The first bullet point is already done in the current release. If I use the following code:

```igorpro
static Function TestWave()
	Make/FREE=1 wv1 = { 15 }
	Make/FREE=1 wv2 = { 17 }
	CHECK_EQUAL_WAVES(wv1, wv2)
End
```

I get the following result:
```
  Assuming equality using mode WAVE_DATA for waves wv1 and wv2; detailed: Waves difference:
Wave1: 0x0851d19340
Wave2: 0x0851d194c0
Dimensions  |Labels|Value    |
------------|------|---------|
[0][0][0][0]|      |15.000000|
            |      |17.000000|
: is false. Assertion "CHECK_EQUAL_WAVES(wv1, wv2)" failed in line 10, procedure "Proc0"
```

The names `wv1` and `wv2` are already included in the output.

t-b:
Good find that NameOfWave already helps us here.

t-b:
@Garados007 On second thought. Can we have the free wave names in

```
Wave1: 0x082651f8b0
Wave2: 0x082651f2d0
```

instead of the memory addresses if these are available?

Garados007:
I think we can have both. Something like this:
`Wave1: wv1 (0x082651f8b0)`
I don't know if the memory address is somewhere needed but this solution will still keep it.

t-b:
@Garados007 Fine with me.

Garados007:
I am trying to solve the trimming of trailing zeros of the numbers. There is a format specifier `%g` that will do this automatically. The current used `%f` will always keep six digits after the decimal point.

This will look like this:
```igorpro
static Function TestWave()
[...]

PR #273, opened 2022-10-12
Enhance Wave Printing

Garados007:
What has been done:
- The names of the waves are always printed
- Only the index from used dimensions are printed. If a wave has only two dimensions so it will print only an index like `[3][1]` and not `[3][1][0][0]`.
- Trailing zeros after the decimal point for FP32 and FP64 waves are removed
- Trailing zeros after the decimal point for Complex FP32 and FP64 waves are removed
- all significant digits of the number are shown. A difference of the number should always be visible.

What has not been done:
- colored string diff. This is impossible to do with the current release of Igor Pro. String Diffing has not been changed.

Closing #221 .

### DIFF
--- a/procedures/unit-testing-utils.ipf
+++ b/procedures/unit-testing-utils.ipf
@@ -14,6 +14,12 @@ static Constant UTF_MAXDIFFCOUNT = 10
 // with two %s and some other text.
 static Constant MAX_STRING_LENGTH = 250
 
+// The number of significant decimal digits a fp32 or fp64 can hold. This value depends on the
+// length of the mantisse (23 bits for fp32 and 53 bits for fp64) and are calculated using this
+// formula: mantisse * log(2). The values are rounded up to the next closest integer.
+static Constant UTF_DECIMAL_DIGITS_FP32 = 7
+static Constant UTF_DECIMAL_DIGITS_FP64 = 16
+
 /// @brief Returns 1 if var is a finite/normal number, 0 otherwise
 ///
 /// @hidecallgraph
@@ -575,10 +581,10 @@ static Function/S SPrintWaveElement(w, row, col, layer, chunk)
 					wfprintf str, "(%d, %d)", wTmpCLS
 				endif
 #endif
+			elseif(minorType & IUTF_WAVETYPE0_FP64)
+				str = "("  + GetNiceStringForNumber(real(w[row][col][layer][chunk]), isDouble=1) + ", " + GetNiceStringForNumber(imag(w[row][col][layer][chunk]), isDouble=1) + ")"
 			else
-				Make/FREE/N=1/C/Y=(WaveType(w)) wTmp
-				wTmp[0] = w[row][col][layer][chunk]
-				wfprintf str, "(%f, %f)", wTmp
+				str = "("  + GetNiceStringForNumber(real(w[row][col][layer][chunk]), isDouble=0) + ", " + GetNiceStringForNumber(imag(w[row][col][layer][chunk]), isDouble=0) + ")"
 			endif
 		else
 			if(minorType & (IUTF_WAVETYPE0_INT8 | IUTF_WAVETYPE0_INT16 | IUTF_WAVETYPE0_INT32))
@@ -597,8 +603,10 @@ static Function/S SPrintWaveElement(w, row, col, layer, chunk)
 					sprintf str, "%d", wLS[row][col][layer][chunk]
 				endif
 #endif
+			elseif(minorType & IUTF_WAVETYPE0_FP64)
+				str = GetNiceStringForNumber(w[row][col][layer][chunk], isDouble=1)
 			else
-				sprintf str, "%f", w[row][col][layer][chunk]
+				str = GetNiceStringForNumber(w[row][col][layer][chunk], isDouble=0)
 			endif
 		endif
 	elseif(majorType == IUTF_WAVETYPE1_TEXT)
@@ -817,6 +825,22 @@ static Function AddValueDiffImpl(table, wv1, wv2, locCount, row, col, layer, chu
 	table[2 * locCount + 1][%ELEMENT] = SPrintWaveElement(wv2, row, col, layer, chunk)
 
 	locCount += 1
+End
+
+// return the string representation of the number with no trailing zeros after the decimal point
+static Function/S GetNiceStringForNumber(n, [isDouble])
+	variable n, isDouble
+
+	variable precision
+	string str
+
+	isDouble = ParamIsDefault(isDouble) ? 0 : !!isDouble;
+
+	precision = isDouble ? UTF_DECIMAL_DIGITS_FP64 : UTF_DECIMAL_DIGITS_FP32
+
+	sprintf str, "%.*g", precision, n
+
+	return str
 End
 
 // This function assumes that v1 != v2 and not both are NaN
--- a/procedures/unit-testing-utils.ipf
+++ b/procedures/unit-testing-utils.ipf
@@ -329,21 +329,34 @@ static Function/S DetermineWaveDataDifference(wv1, wv2, tol)
 
 	string msg
 	variable isComplex1, isComplex2
-	string wvName1, wvName2
+	string wvId1, wvId2, wvName1, wvName2
 	string wvNamePrefix, tmpStr1, tmpStr2
 
 	// Generate names for reference
 	if(WaveExists(wv1))
-		wvName1 = GetWaveNameInDFStr(wv1)
+		wvId1 = GetWaveNameInDFStr(wv1)
+		wvName1 = NameOfWave(wv1)
 	else
+		wvId1 = "_null_"
 		wvName1 = "_null_"
 	endif
 	if(WaveExists(wv2))
-		wvName2 = GetWaveNameInDFStr(wv2)
+		wvId2 = GetWaveNameInDFStr(wv2)
+		wvName2 = NameOfWave(wv2)
 	else
+		wvId2 = "_null_"
 		wvName2 = "_null_"
 	endif
-	wvNamePrefix = "Wave1: " + wvName1 + "\rWave2: " + wvName2 + "\r"
+	if(strsearch(wvId1, " in ", 0) >= 0)
+		sprintf wvNamePrefix, "Wave1: %s\r", wvId1
+	else
+		sprintf wvNamePrefix, "Wave1: %s (%s)\r", wvName1, wvId1
+	endif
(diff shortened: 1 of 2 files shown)


t-b:
Review:

ddd02617 (Show significant dimensions in wave diff, 2022-10-11)

commit message:
> In the wave diff window are normally all four dimensions shown even if
> they are not used or doesn't exists. With this change it will only
> render the number of dimensions (up to four) that are for the wave

What is the wave diff window? The history?
What is the difference between "not used" and "doesn't exists"? The dimension of a wave is a property like it's type (fp32, int, etc.). So it either has columns or not.

The following

```igorpro
		case 3:
			sprintf str, "[%d][%d][%d]", row, col, layer
			break
		default:
			sprintf str, "[%d][%d][%d][%d]", row, col, layer, chunk
			break
```

is not nice. The function `AddValueDiffImpl` now returns the same results for `dimensions == 4` and `dimensions == 5`. I think it would be better to handle 4 explicitly and abort in the default branch. EvaluateRTE handles funcType in that way.

853f7448 (Show wave names in diff temp wave output, 2022-10-11)

Functionality wise the new output looks a bit odd for permanent waves:

```igorpro

Function Test()

	Make/FREE=1 wv1 = {1}
	Make/FREE=1 wv2 = {2}

	CHECK_EQUAL_WAVES(wv1, wv2)

	Make/O wv3 = {3}
	Make/O wv4 = {4}

	CHECK_EQUAL_WAVES(wv3, wv4)

	CHECK_EQUAL_WAVES(wv1, wv4)
End
```

gives

```
•runtest("procedure")
  Start of test "Unnamed"
  Entering test case "Test"
  Assuming equality using mode WAVE_DATA for waves wv1 and wv2; detailed: Waves difference:
Wave1: wv1 (0x0869a9150)
Wave2: wv2 (0x0869a89b0)
Dimensions|Labels|Value|
----------|------|-----|
[0]       |      |1    |
          |      |2    |
: is false. Assertion "CHECK_EQUAL_WAVES(wv1, wv2)" failed in line 12, procedure "Procedure"
  Assuming equality using mode WAVE_DATA for waves wv3 and wv4; detailed: Waves difference:
Wave1: wv3 (wv3 in root:tempFolder0:)
Wave2: wv4 (wv4 in root:tempFolder0:)
Dimensions|Labels|Value|
----------|------|-----|
[0]       |      |3    |
          |      |4    |
: is false. Assertion "CHECK_EQUAL_WAVES(wv3, wv4)" failed in line 17, procedure "Procedure"
  Assuming equality using mode WAVE_DATA for waves wv1 and wv4; detailed: Waves difference:
Wave1: wv1 (0x0869a9150)
Wave2: wv4 (wv4 in root:tempFolder0:)
Dimensions|Labels|Value|
----------|------|-----|
[0]       |      |1    |
          |      |4    |
: is false. Assertion "CHECK_EQUAL_WAVES(wv1, wv4)" failed in line 19, procedure "Procedure"
  Leaving test case "Test"
  Failed with 3 errors
  Test finished with 3 errors
  End of test "Unnamed"
```

50a2d288 (Nice number formating for double wave diffs, 2022-10-11)

Commit message:

```
Before only 6 digits after the decimal point are shown.
```

are -> were

Code style:
- No trailing whitespace
- No space after if/for etc.
- Prefer strsearch over stringmatch
- `CmpStr(A) != 0` -> `!CmpStr(A)`
- for loop variables are (from outer to inner): i, j, k, l

```
isDouble = ParamIsDefault(isDouble) ? 0 : isDouble != 0;
```

The more idiomatic approach is to use `!!isDouble` instead of `isDouble != 0`.

```
	precission = isDouble ? 15.995 : 6.93
```

These are magic numbers. It usually best to introduce constants for these. And in this case also explain (commit message or comment both good) how these are calculated.

precission -> precision 

The format string building is not required as sprintf support dynamic field width/precision.

From the help:

> You can replace both the field width and precision specifiers with an asterisk. This 
> tells printf to get the field width or precision specifier from a parameter. For 
> example:
> printf "%*.*f\r" 4, 3, 1e6*PI

```
			part = str[right, right]
```

I think `part = str[right]` does the same or?

Can you also add some units tests for GetNiceStringForNumber? You should be easily reach 100% coverage ;)

Garados007:
Where I have to put the test case? In `Various.ipf` or somewhere else?

t-b:
> Various.ipf

Yes this is a good spot for now.

Garados007:
As I am writing some test cases I noticed that the string representation for really big numbers are a little bit odd. For example `1e30` will result in `1000000000000000019884624838656`.

For this I have the following solutions:
1. If the number gets larger than lets say 1e16 we will use the scientific notation `1e30` and show all digits before the `e`.
2. Float32 can only show 6, Float64 16, decimal digits. All digits behind will be set to `0`. So `1e30` will be shown as `1000000000000000000000000000000`.

t-b:
Solution 1 sounds attractive. But the "lets say 1e16" part should be justified and not just a random pick. @MichaelHuth What do you think?

PS: Solution 2 is a big NO-NO.

t-b:
This is also not an IP bug, see https://godbolt.org/z/7eeEb4M57.

Garados007:
I think 16 is for FP64 a good number. The mantissa for FP64 is 53 bits and that is converted to decimal approximate 16 digits (53 * log(2)). So the maximum precision for FP64 is 16 decimal digits. If I want to display a number that's larger than 1e17 I will have a digit before the decimal point that wasn't included in the FP64 mantissa. So I think 1e16 is a good cutoff to start display scientific numbers.

FP32 has the same argumentation can display 6 decimal digits (23 mantissa * log(2)).

Garados007:
Very small numbers (for example 1e-30) do not have this problem as they are displayed correctly, but these can be difficult to capture visually. 1e-30 would be `0.000000000000000000000000000001`.  There are lots of zeros after the decimal point until the first significant digit. I think we should also limit this in this direction.

MichaelHuth:
I think:
The main goal is to get two numbers displayed that are at some digit different in a way that allows to identify the difference easily.

Thus, there are multiple points:
- numbers can get quite small or large, so boiling down to significant digits is one step to reduce the printout. For a specific number this results on two offsets: start and end of significant digits. After some thought I would also round the 53 * log(2) up to the next higher integer, because we already know that the two numbers to compare are different.
- The displayed values must not be modified and thus wrong, like replacing digits with zero.
- The last comment refers to an actual display problem, so for 1E-30 the start digit offset is -30 and the end offset -46. The display with leading zeroes is obviously impractical. Same applies for very large numbers.

Thus, I would differentiate the two possible cases:
1. The start digit offset is different
2. The start digit offset is the same and a digit in the overlapping digit range is different

For case 1 a display with a single digit and exponential display is sufficient to see a difference, e.g.
```
1E30
1E-30
```
For case 2 the exponents are basically the same. Thus, I would at least display from start offset digit to end offset digit of the smallest type of both numbers.  If the distance is now more than e.g. 20 digits to the dot I would display the exponential notation, where the exponent is the same for both numbers otherwise without exponent. And make sure that the dot shows in the same position when both numbers are printed above each other.

The NaN/Inf/-Inf cases are trivial.

Add for case 1: The distance to the dot should also be considered there and it should be only switched to exponential notation (for both) of one number is too far off.

With that approach we can catch big number differences and close number differences and have a nice diff display for both.

Is that idea reasonable? Improvements on that?

MichaelHuth:
@Garados007 
I chatted with Thomas about this and we came to the conclusion that a proper solution is quite complex. (more complex than initially thought)
There are also others who had thought about that and provide a solution, that would be e.g. a Igor Pro extension in the future.
( https://github.com/jk-jeon/dragonbox )

To not escalate the complexity further here, we agreed that we could live with a solution to print the numbers as "%.16g" for floating point numbers.

Garados007:
Ok. I started to implement [your comment](https://github.com/byte-physics/igor-unit-testing-framework/pull/273#issuecomment-1278248801) right now including the big complexity.

Should I remove 33f6c7ae3097a99f181ce2141f63610391b6f47c and only use `%.16f` for all numbers (FP32 & FP64) or `%.7f` for FP32 and `%.16f` for FP64? Do we need the elimination of trailing zeros?

Garados007:
sorry I pushed my unfinished and unrebased stuff. Just ignore it.

t-b:
@Garados007 Am I assigned correctly? As stated in https://github.com/byte-physics/igor-unit-testing-framework/pull/273#issuecomment-1279324483 we think %g is the way to go. Removal of trailing zeros would still be nice.

t-b:
33f6c7ae (Nice number formating for double wave diffs, 2022-10-11)

No space after switch.

d889a2e1 (Show wave names in diff temp wave output, 2022-10-11)

Needs more work see below.

ab5b29ba (Show significant dimensions in wave diff, 2022-10-11)

Good!

Can we have a new PR for only the fixed version of 33f6c7ae and ab5b29ba? In that way we can fasttrack these changes.

Regarding the wave name output. 

- The strsearch approach with ` in ` is very hacky
- You are duplicating the work with calling NameOfWave again as GetWaveNameInDFStr() does that as well. Can't we have one routine which generates 

```
sprintf wvNamePrefix, "%s (%s)\r", wvName1, wvId1
```

and then everyone can use that? Per our chat today this should then always be used instead of NameOfWave in the other places. This then also needs some unit tests. Thanks.

```
$ git grep --show-function NameOfWave main
main:procedures/unit-testing-assertion-wrappers.ipf=static Function EQUAL_WAVE_WRAPPER(wv1, wv2, flags, [mode, tol])
main:procedures/unit-testing-assertion-wrappers.ipf:     sprintf str, "Assuming equality using mode %s for waves %s and %s", EqualWavesModeToString(mode), NameOfWave(wv1), NameOfWave(wv2)
main:procedures/unit-testing-utils.ipf=static Function/S GetWaveNameInDFStr(w)
main:procedures/unit-testing-utils.ipf:  str = NameOfWave(w)
```

Garados007:
ab5b29ba97c8f0510555abbb754fdd4b2deaeb43 is now tracked in PR  #281.

Garados007:
b42a3b8bff6b22cf3551456fd8c141d6772b65a1 is now tracked in PR #282

Garados007:
d889a2e162da76ba10df57bfe36cccd17dcda763 is now tracked in PR #283

Garados007:
All parts of this PR are now tracked in their own PR. I think we can close this PR right now.

t-b:
> All parts of this PR are now tracked in their own PR. I think we can close this PR right now.

Sure!

PR #278, opened 2022-10-18
Enable fixed log file names

### LINKED ISSUE #38: automatic log file naming
I think it would be good to have the log files produced with a static name instead of a name with the date.

Three reasons:
* automatic processing of the log file is more convenient if I know the exact name. Otherwise The directory needs to be cleaned of *.log and the single *.log file is the output of the script.
* Keeping the log file open with a text editor would automatically gives the newly created output
* a tailing pointer would be possible for use i.e. in `tail -f`

If data overwrite is a problem, a log rotate would be a possible solution.

Garados007:
RunTest got a new optional argument fixLogName for which the output log file get a predictable name. The new file name will be IUTF_Test.log. If this flag is not set the naming will be unchanged and contains the current timestamp.

Close #38.

### DIFF
--- a/procedures/unit-testing-autorun.ipf
+++ b/procedures/unit-testing-autorun.ipf
@@ -100,6 +100,7 @@ End
 Function/S GetBaseFilename()
 	dfref dfr = GetPackageFolder()
 	SVAR/Z/SDFR=dfr baseFilename
+	SVAR/Z/SDFR=dfr baseFilenameOverwrite
 
 	if(!SVAR_Exists(baseFilename))
 		string/G dfr:baseFilename = ""
@@ -109,7 +110,13 @@ Function/S GetBaseFilename()
 	if(strlen(baseFilename))
 		return baseFilename
 	endif
-	sprintf baseFilename, "%s_%s_%s", IgorInfo(1), Secs2Date(DateTime, -2), ReplaceString(":", Secs2Time(DateTime, 3), "-")
+
+	if(SVAR_Exists(baseFilenameOverwrite) && strlen(baseFilenameOverwrite))
+		baseFilename = baseFilenameOverwrite
+	else
+		sprintf baseFilename, "%s_%s_%s", IgorInfo(1), Secs2Date(DateTime, -2), ReplaceString(":", Secs2Time(DateTime, 3), "-")
+	endif
+
 	return baseFilename
 End
 
--- a/procedures/unit-testing-autorun.ipf
+++ b/procedures/unit-testing-autorun.ipf
@@ -92,13 +92,13 @@ End
 
 /// resets a global filename template string for output
 Function ClearBaseFilename()
-	dfref dfr = GetPackageFolder()
+	DFREF dfr = GetPackageFolder()
 	string/G dfr:baseFilename = ""
 End
 
 /// creates a new filename template, if template already present return current
 Function/S GetBaseFilename()
-	dfref dfr = GetPackageFolder()
+	DFREF dfr = GetPackageFolder()
 	SVAR/Z/SDFR=dfr baseFilename
 	SVAR/Z/SDFR=dfr baseFilenameOverwrite
 
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -63,6 +63,8 @@ Constant IUTF_DEBUG_NVAR_SVAR_WAVE = 0x04
 Constant IUTF_DEBUG_FAILED_ASSERTION = 0x08
 /// @}
 
+static StrConstant FIXED_LOG_FILENAME = "IUTF_Test"
+
 StrConstant IUTF_TRACE_REENTRY_KEYWORD = " *** REENTRY ***"
 
 static Constant TEST_CASE_TYPE = 0x01
@@ -3404,12 +3406,18 @@ End
 ///                         HTMLCREATION:boolean When set to zero, no htm result files are created at the end of the run
 ///                         REGEXP:boolean When set, traceWinList is interpreted as regular expression
 ///
+/// @param   fixLogName     (optional) default 0 disabled, enabled when set to 1: @n
+///	                        If enabled the output files that will be generated after an autorun will have predictable names like
+///                         "IUTF_Test.log". If disabled the file names will always contain the name of the procedure file and a
+///                         timestamp.
+///
 /// @return                 total number of errors
-Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp, allowDebug, debugMode, keepDataFolder, traceWinList, traceOptions])
+Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp, allowDebug, debugMode, keepDataFolder, traceWinList, traceOptions, fixLogName])
 	string procWinList, name, testCase
 	variable enableJU, enableTAP, enableRegExp
 	variable allowDebug, debugMode, keepDataFolder
 	string traceWinList, traceOptions
+	variable fixLogName
 
 	// All variables that are needed to keep the local function state are wrapped in s
 	// new var/str must be added to strRunTest and added in SaveState/RestoreState functions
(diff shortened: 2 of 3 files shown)


Garados007:
I want to check if the limitation that the experiment has to be saved is removed before I mark this PR as ready.

Garados007:
> I want to check if the limitation that the experiment has to be saved is removed before I mark this PR as ready.

I added a new check that will allow unsaved experiments when certain features are not used.

Garados007:
Just a rebase of the branch with no changes.

procedures/unit-testing-basics.ipf
```diff
@@ -3464,11 +3467,23 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
...
 		CreateHistoryLog()
 
-		PathInfo home
-		if(!V_flag)
-			sprintf msg, "Error: Please Save experiment first."
-			UTF_PrintStatusMessage(msg)
-			return NaN
+		// Check if the current experiment needs to be saved. This is only required if certain features
+		// of UTF are used.
+		reqSave = 0
```
MichaelHuth, line 3472: I think you don't need the reqSave variable here. You can combine the conditions in the if.

procedures/unit-testing-autorun.ipf
```diff
@@ -109,7 +110,13 @@ Function/S GetBaseFilename()
 	if(strlen(baseFilename))
 		return baseFilename
 	endif
-	sprintf baseFilename, "%s_%s_%s", IgorInfo(1), Secs2Date(DateTime, -2), ReplaceString(":", Secs2Time(DateTime, 3), "-")
+
+	if (SVAR_EXISTS(baseFilenameOverwrite))
```
MichaelHuth, line 114: I think the approach to remove the global string when the option is not set has the disadvantage that for that case the full current state of the IUTF is not visible in its datafolder.

How about creating baseFilenameOverwrite always and use an empty string as disabled. You can check that with UTF_Utils#IsEmpty here.

btw: The there is a space here after the if and the function used here is SVAR_Exists.

procedures/unit-testing-basics.ipf
```diff
@@ -458,7 +458,7 @@ End
 Function EvaluateResults(result, str, flags)
 	variable result, flags
 	string str
-
```
MichaelHuth, line 461: Please put such unrelated whitespace corrections in a separate commit.

procedures/unit-testing-basics.ipf
```diff
@@ -3447,6 +3453,13 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
 
 	else
 		// no early return/abort above this point
+		if(!!fixLogName)
+			dfref dfr = GetPackageFolder()
```
MichaelHuth, line 3460: While there are some places here where we write "dfref" the current convention is "DFREF".
This is related to be consistent with other reference assignments such as WAVE and FUNCREF for example.

procedures/unit-testing-basics.ipf
```diff
@@ -3447,6 +3453,13 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
 
 	else
 		// no early return/abort above this point
+		if(!!fixLogName)
```
MichaelHuth, line 3459: Please use a separate line before to properly initialize fixLogName.
Like:
`fixLogName = ParamIsDefault(fixLogName) ? 0 : !!fixLogName`

This way it is easy to see what the default value of the optional parameter should be.
Then please also add that value to the fixLogName argument comment description above. (default 0 disabled)

procedures/unit-testing-basics.ipf
```diff
@@ -3447,6 +3453,13 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
...
+			dfref dfr = GetPackageFolder()
+			string/G dfr:baseFilenameOverwrite = "IUTF_Test"
+		else
+			dfref dfr = GetPackageFolder()
+			KillVariables/Z dfr:baseFilenameOverwrite
```
MichaelHuth, line 3464: See my comment above. Make sure you have only a single location where you create baseFilenameOverwrite through
`string/G dfr:baseFilenameOverwrite`

procedures/unit-testing-basics.ipf
```diff
@@ -3447,6 +3453,13 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
 
 	else
 		// no early return/abort above this point
+		if(!!fixLogName)
+			dfref dfr = GetPackageFolder()
+			string/G dfr:baseFilenameOverwrite = "IUTF_Test"
```
MichaelHuth, line 3461: Please create a StrConstant for the fixed log file name.

MichaelHuth (changes requested):
General approach is good.

See comments in the code.

Please also read the commit message once more, e.g. "The new new will be IUTF_Test.log."  -> "The fixed log file name is IUTF_Test.log".

"Before any experiment has to be saved somewhere on the disk" ->
"Before any experiment had to be saved on the disk"

"This was required because RunTest
checked if a local "home" directory of the experiment exists. This was
done because J-Unit, TAP and tracing need a location to store or
manipulate their data."
->
"This was a requirement because the IUTF needs a set home path to save data for JUnit, TAP and tracing."

procedures/unit-testing-basics.ipf
Garados007: The problem here is that the if statement become very large and less understandable. It would help if I can wrap the code over multiple lines but this is not available in Igor Pro 6.

t-b: Or maybe move the decision logif for reqSave into a function? This function can have multiple if/elseif statements with early return.

Garados007: During refactoring to get a separate function I noticed there is a better and shorter approach.

procedures/unit-testing-basics.ipf
Garados007: A simple search returned me 28 examples with "dfref". Maybe we should correct that in the future.

Garados007: I fixed this in 05c5dbb2587a94577124a55c0f05b4f89d73b7a6. If it is preferred I can create a new issue and add this commit to there.

procedures/unit-testing-autorun.ipf
```diff
@@ -109,7 +110,13 @@ Function/S GetBaseFilename()
 	if(strlen(baseFilename))
 		return baseFilename
 	endif
-	sprintf baseFilename, "%s_%s_%s", IgorInfo(1), Secs2Date(DateTime, -2), ReplaceString(":", Secs2Time(DateTime, 3), "-")
+
+	if(SVAR_Exists(baseFilenameOverwrite) && strlen(baseFilenameOverwrite))
```
MichaelHuth, line 114: I know its intriguing to just copy the strlen check from above, though we use now UTF_Utils#IsEmpty(str) for this.
The advantage of UTF_Utils#IsEmpty is that it also checks for a null string, where strlen would cause an RTE if you can resolve the baseFileNameOverwrite global and baseFileNameOverwrite is a null string.

MichaelHuth (changes requested):
Use UTF_Utils#IsEmpty, see comment

procedures/unit-testing-autorun.ipf
Garados007: I created #303 to target the `strlen()` issue in general. The requested place will be fixed of course.

Garados007:
I removed 98615ed29ce9c5e91c58a31bf2e7e3de1abe5e84 (remove trailing whitespaces, 2022-10-30) from this PR because it is already contained in main.

procedures/unit-testing-autorun.ipf
Garados007: For documentation: I have to revert to `strlen` at this place. `baseFilenameOverwrite` is an `SVAR` and therefore cannot be null and `UTF_Utils#IsEmpty` is only accepting strings.

PR #279, opened 2022-10-19
unit-testing-junit.ipf: Improve accuracy of test case/suite durations

t-b:
Ever since the introduction of JUNIT support in 7b71266a (Added JUNIT support, 2017-03-16) we have calculated the durations of a test case or test suite using DateTime.

The default choice for relative time tracking is usually stopmstimer(-2).

And this can also be verified using the following example:

print/D stopmstimer(-2)/1000, stopmstimer(-2)/1000, stopmstimer(-2)/1000, stopmstimer(-2)/1000, stopmstimer(-2)/1000
  11962615.9893  11962616.1171  11962616.2341  11962616.3487  11962616.4618
print/D datetime , datetime, datetime, datetime, datetime
  3749025804.137  3749025804.137  3749025804.137  3749025804.138  3749025804.138

Here the timestamps for stopmstimer vary with every call, but DateTime stays constant.

So let's use stopMSTimer(-2) for higher accuracy.

This can be verified executing the MIES [1] test case WorksWithSameWordAndRepl which now has an execution time of 0.006 s instead of 0.000 s.

[1]: https://github.com/AllenInstitute/MIES

### DIFF
--- a/procedures/unit-testing-junit.ipf
+++ b/procedures/unit-testing-junit.ipf
@@ -377,7 +377,7 @@ Function JU_TestSuiteBegin(s, name, procWin)
 	s.juTS.timestamp = JU_GetISO8601TimeStamp()
 	s.juTS.hostname = "localhost"
 	s.juTS.tests = s.testCaseCount
-	s.juTS.timeStart = DateTime
+	s.juTS.timeStart = JU_GetRelativeTime()
 	s.juTS.failures = 0
 	s.juTS.errors = 0
 	s.juTS.skipped = 0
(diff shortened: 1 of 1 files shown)


Garados007 (approved):
LGTM

MichaelHuth (approved):
LGTM, though one comment on that:

the timeTaken XML field is of type xs:decimal and is currently created with a ms precision through %.3f.
Would it make sense to increase the precision here since we got more precise time differences now?

(for xs:decimal scientific notation is not valid)

PR #286, opened 2022-10-24
AnalyzeTracingResult: Handle missing INSTRUDATA_FILENAME really grace…

### LINKED ISSUE #285: LoadWave: Missing file dialog breaks test run
grafik

t-b:
…fully

The operation LoadWave does not have a way to prevent a "Missing file dialog". So we have to manually check that the file exists before. Bug introduced in de9e739c (Multiple changes and fixes, 2021-07-15).

This can happen if the experiment is resaved during test run to a different folder, i.e. the path home changes.

Close #285.


Garados007 (approved):
LGFM

PR #288, opened 2022-10-25
Enhance error reporting in CHECK_WAVE

### LINKED ISSUE #220: Enhance error reporting for CHECK_WAVE
```igorpro
Function Test()

CHECK_WAVE($"", NUMERIC_WAVE)
End
```

gives

```
•runtest("Procedure", testcase = "Test")
  Start of test "Unnamed"
  Entering test case "Test"
  Assumption that the wave's main type is 1: is false. Assertion "CHECK_WAVE($"", NUMERIC_WAVE)" failed in line 31, procedure "Procedure"
  Leaving test case "Test"
  Failed with 1 errors
  Test finished with 1 errors
  End of test "Unnamed"
```

which leaves the reader a bit clueless.

I would think we want:
- strings instead of numeric values in the output message (prefer NUMERIC_WAVE over 1)
- output the actual type of the wave

Garados007:
Makes the output of the wave type more user understandable. This shows the specific type as a string which contains the same names as the flags that are used to specify the type in CHECK_WAVE.

Close #220.

### DIFF
--- a/procedures/unit-testing-assertion-checks.ipf
+++ b/procedures/unit-testing-assertion-checks.ipf
@@ -272,14 +272,22 @@ static Function AreWavesEqual(wv1, wv2, mode, tol, detailedMsg)
 	return result
 End
 
-static Function HasWaveMajorType(wv, majorType)
+static Function AddIfFlagSet(var, flag, flagString, str)
+	variable var, flag
+	string flagString
+	string &str
+
+	if((var & flag) == flag)
+		str = AddListItem(flagString, str, ", ")
+	endif
+End
+
+static Function GetWaveMajorType(wv)
 	WAVE/Z wv
-	variable majorType
 
-	variable type, type1, type2
-
-	type2 = WaveType(wv, 2)
-	type1 = WaveType(wv, 1)
+	variable type = 0
+	variable type2 = WaveType(wv, 2)
+	variable type1 = WaveType(wv, 1)
 
 	if(type1 > 0 && type1 <= 4)
 		type = type | 2^(type1 - 1)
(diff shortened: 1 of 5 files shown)


Garados007:
Just a rebase of the branch with no changes.

t-b:
While testing I found #296, I'll let you be the judge if this should be done before this PR or afterwards.

Review:

-

```
	WAVE/Z wv
	CHECK_WAVE(wv, NULL_WAVE, minortype = UNSIGNED_WAVE)
```

gives

```
  Expect wave's sub type to be UNSIGNED_WAVE but got 4096: is false. Assertion "CHECK_WAVE(wv, NULL_WAVE, minortype = UNSIGNED_WAVE)" failed in line 12, procedure "TestWaveType.ipf"
```

and I would not have expected a number here. Maybe we should always fail if the wave is null and a minor type was requested? Separate issue/PR?

-

```
+    if(!CmpStr(str, ""))
+        str = num2str(type)
+    endif
```

should return early.

-

```
+    if(type == 0)
+        return "NON_NUMERIC_WAVE"
+    endif

```

Good catch. But doesn't that mean we should have a NON_NUMERIC_WAVE constant as well?

- While the output is mostly complete and readable how about something like

```
Expect wave's main type to be 'NULL_WAVE' but got 'FREE_WAVE, NUMERIC_WAVE': 
```

with single quotes and `, ` instead of `;` as I think this reads better. Or what do you think?

- I think the new pass-by-reference parameters to HasWaveMajorType/HasWaveMinorType does the job. But makes the call sites, especially the tests, quite ugly. How about extracting a function like GetWaveMajorType/GetWaveMinorType out. This function can then be used for getting the type for pretty-printing. In that way you don't need to touch the HasWaveMajorType/HasWaveMinorType tests.

- I don't think it makes sense to combine NULL_WAVE with anything else in the tests. Likewise GetWaveMajorTypeString should probably early return with NULL_WAVE?

- I tend to think that having a small function like

````
Function AddIfSet(type, flag, flagAsString, str)
   variable type, flag
   string flagAsString
   string &str

	if((type & flag) == flag)
		str = AddListItem(flagAsString, str)
	endif
End
```

would help readability for the stringify getters. But I'll let you decide.

Garados007:
> - While the output is mostly complete and readable how about something like
>     ```
>     Expect wave's main type to be 'NULL_WAVE' but got 'FREE_WAVE, NUMERIC_WAVE':
>     ```
>     with single quotes and , instead of ; as I think this reads better. Or what do you think?

I support this idea. I also thought that the current approach doesn't look nice but it was a quick and easy solution. ;)

Garados007:
> -  ```igorpro
>   	WAVE/Z wv
>   	CHECK_WAVE(wv, NULL_WAVE, minortype = UNSIGNED_WAVE)
>    ```
>     gives
>     ```
>       Expect wave's sub type to be UNSIGNED_WAVE but got 4096: is false. Assertion "CHECK_WAVE(wv, NULL_WAVE, minortype = UNSIGNED_WAVE)" failed in line 12, procedure "TestWaveType.ipf"
>     ```
>     and I would not have expected a number here. Maybe we should always fail if the wave is null and a minor type was requested? Separate issue/PR?

I think I can include this in the current PR. An interesting note is, that 4096 (NULL_WAVE) should be supported by the minor type. I thought that is only the case for the major type.

Garados007:
I applied your suggestions and included #296.

t-b:
Review:

53ea29bf (Require valid types in CHECK_WAVE, 2022-11-01)

This is good to go. Thanks. If you create a new PR for the next commit we merge this one.

c8ad3acd (Enhance error reporting in CHECK_WAVE, 2022-10-25)

We never `Abort` from a `*_WRAPPER` function. Have a look at EQUAL_WAVE_WRAPPER how it handles invalid input with `EvaluateResults(0,...`. EQUAL_WAVE_WRAPPER does also check that only existing bits are passed for `mode`. I would expect the same check for minor and major wave type. I'd say you don't need to check that it makes sense to combine bit X with bit Y, only that X and Y are by themselves valid.

t-b:
@Garados007 Nice!

t-b (approved):

PR #294, opened 2022-10-29
Make string comparisons always case sensitive by default

### LINKED ISSUE #291: Make the default for string matching case-sensitive.
For waves comparison the string matching is always case-sensitive and byte-wise for Igor >= 7.05. For CHECK_EQUAL_STR the default for string matching is case-insensitive which is not intuitive.

In my opinion we should change it to case-sensitive (with the option for case-insensitive matching) for all string checks because in a test case we are expecting exact data and we are not working with user input.

t-b:
- [x] Add VTTE test which tests the *default* value for case_sensitive.

Close #291

### DIFF
--- a/docu/examples/example2-plain.ipf
+++ b/docu/examples/example2-plain.ipf
@@ -23,8 +23,9 @@ static Function VerifyStringComparison()
 	string strLow      = "123abc"
 	string strUP       = "123ABC"
 
-	// by default string comparison is done case insensitive
-	CHECK_EQUAL_STR(strLow, strUP)
+	// by default string comparison is done case sensitive
+	// so the following would fail with CHECK/REQUIRE
+	WARN_EQUAL_STR(strLow, strUP)
 	// It can be specificylly enabled or disabled.
 	CHECK_EQUAL_STR(strLow, strUP, case_sensitive = 0)
 	// Now we use WARN because the two strings are not equal.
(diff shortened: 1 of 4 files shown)


t-b:
We've got 4 assertion failures in MIES which has 1332 testcases. So I'd say we can flip the default.

Garados007:
4 failures? I thought there would be way more.

I support this PR and LGTM!

Garados007 (approved):

MichaelHuth (changes requested):
Code wise thats ok. But I think in the commit message the last sentence should refer to flipping the default in EQUAL_STR_WRAPPER and not EQUAL_WAVE_WRAPPER ?

t-b:
@MichaelHuth Good point. Fixed.

Here is the range diff:
```
$ git range-diff @{u}...HEAD
1:  3274264 ! 1:  89c28de EQUAL_STR_WRAPPER: Make the default case sensitive
    @@ Commit message

         We can't flip the case sensitivity switch for EqualWaves as there is none.
         And a testing framework should also be as strict as possible, so we flip
    -    the default for EQUAL_WAVE_WRAPPER to case sensitive.
    +    the default for EQUAL_STR_WRAPPER to case sensitive.

         Reported-by: @Garados007
         Test-code-by: @Garados007
```

PR #295, opened 2022-10-30
Abort test run if RTE ocurred in UTF

### LINKED ISSUE #264: Check RTE condition before CallTestCase and abort gracefully
In case the IUTF raised a RTE we silently clear that before CallTestCase.

To prevent possible unexpected behavior, we should check against pending RTEs before changing to user code.
In case of RTE abort the IUTF run as the current IUTF state is inconsistent then.

Garados007:
Abort the test run if some unhandled RTE still exists before CallTestCase(). These errors are most likely a bug in UTF and leave the UTF in an inconsistent state. It is more safe to stop the run at that moment instead of continuing it.

Before this change any RTE that could happen before the run of the test case are just ignored and thrown away.

Close #264.

### DIFF
--- a/internal_dev/IM_tap_example1a-uncaught-aborts.ipf
+++ b/internal_dev/IM_tap_example1a-uncaught-aborts.ipf
@@ -38,5 +38,5 @@ End
 // #TAPDescription: Planned abort here
 static Function Bail_Out()
 
-	AbortNow()
+	Abort
 End
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -442,7 +442,7 @@ static Function DebugOutput(str, booleanValue)
 
 	str = str + ": is " + SelectString(booleanValue, "false", "true") + "."
 	if(EnabledDebug())
-		UTF_PrintStatusMessage(str)
+		ReportError(str, incrErrorCounter = 0)
 	endif
 End
 
@@ -457,6 +457,12 @@ Function SetTestStatusAndDebug(str, booleanValue)
 	SetTestStatus(str)
 End
 
+/// Evaluate the result of an assertion that was used in a testcase. For evaluating internal errors use
+/// ReportError* functions.
+/// @param result          Set to 0 to signal an error. Any value different to 0 will be considered as success.
+/// @param str             The message to report.
+/// @param flags           A combination flags that are used by ReportResults() to determine what to do if result
+///                        is in an error state.
 Function EvaluateResults(result, str, flags)
 	variable result, flags
 	string str
@@ -684,9 +690,8 @@ Function PrintFailInfo([prefix])
 	str = getInfo(0)
 	message = prefix + status + " " + str
 
-	UTF_PrintStatusMessage(message)
 	type = "FAIL"
-	UTF_ToSystemErrorStream(message)
+	ReportError(message, incrErrorCounter = 0)
 
 	if(TAP_IsOutputEnabled())
 		SVAR/SDFR=dfr tap_diagnostic
@@ -710,7 +715,8 @@ static Function setAbortFlag()
 	variable/G dfr:abortFlag = 1
 End
 
-/// @brief Wrapper function result reporting
+/// @brief Wrapper function result reporting. This functions should only be called for
+///        assertions in user test cases. For internal errors use ReportError* functions.
 ///
 /// @param result Return value of a check function from `unit-testing-assertion-checks.ipf`
 /// @param str    Message string
@@ -734,15 +740,11 @@ static Function ReportResults(result, str, flags)
 				incrError()
 			endif
 			if(flags & ABORT_FUNCTION)
-				abortNow()
+				setAbortFlag()
+				Abort
 			endif
 		endif
 	endif
-End
-
-Function abortNow()
-	setAbortFlag()
-	Abort
 End
 
 /// Resets the abort flag
@@ -929,13 +931,13 @@ static Function abortWithInvalidHooks(hooks)
 	numEntries = DimSize(info, 0)
 	for(i = 0; i < numEntries; i += 1)
 		if(NumberByKey("N_PARAMS", info[i]) != 1 || NumberByKey("N_OPT_PARAMS", info[i]) != 0 || NumberByKey("PARAM_0_TYPE", info[i]) != 0x2000)
-			sprintf msg, "The override test hook \"%s\" must accept exactly one string parameter.\r", StringByKey("NAME", info[i])
-			Abort msg
+			sprintf msg, "The override test hook \"%s\" must accept exactly one string parameter.", StringByKey("NAME", info[i])
+			ReportErrorAndAbort(msg)
 		endif
 
 		if(NumberByKey("RETURNTYPE", info[i]) != 0x4)
-			sprintf msg, "The override test hook \"%s\" must return a numeric variable.\r", StringByKey("NAME", info[i])
-			Abort msg
+			sprintf msg, "The override test hook \"%s\" must return a numeric variable.", StringByKey("NAME", info[i])
+			ReportErrorAndAbort(msg)
 		endif
 	endfor
 End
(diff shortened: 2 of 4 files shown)


t-b:
Review:

Your code works, and its even a creative use of AbortONRTE. But I'd suggest a slightly different implementation.
- Query the RTE state with GetRTError() passing `0` only queries it, whereas `1` also clears it
- Returning immediately is not the right thing I think. Although you are right that we are in a problematic state as something went wrong, we still can try to finish as usual in the error case. CallTestCase does the triumvirate of

```
					sprintf msg, "Fatal: Reentry function %s does not meet required format for Complex argument.", func
					UTF_PrintStatusMessage(msg)
					incrError()
					abortNow()
```

which actually should be wrapped in a function. That works I think better. Or maybe like AfterTestCase does it.

- Don't introduce a message type like "CRTITAL" adhoc. This would need a planned and documented appraoch.

Garados007:
If I introduce the following error in UTF:
```diff
diff --git a/procedures/unit-testing-basics.ipf b/procedures/unit-testing-basics.ipf
index 1579bd5..5781866 100644
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -3688,6 +3688,9 @@ Function RunTest(procWinList, [name, testCase, enableJU, enableTAP, enableRegExp
 
 			if(!skip)
 
+				Wave test = $""
+				variable testvar = test[0]
+
 				if(GetRTError(0))
 					message = GetRTErrMessage()
 					err = GetRTError(1)
```

I get the following output:
```
  Start of test "Unnamed"
  Entering test suite "Proc0"
  Entering test case "mymodule#Test4"
  Fatal: Internal runtime error in UTF 330:"a wave read;Attempt to operate on a null (missing) wave" before executing test case "mymodule#Test4".
  Leaving test case "mymodule#Test4"
  Failed with 1 errors
  Leaving test suite "Proc0"
  Test finished with 1 errors
  End of test "Unnamed"
```

t-b:
Review:
015108aa (Simplify the report of internal errors, 2022-11-03)

This looks much nicer! Would it make sense to convert even more ad-hoc error reporting to the new infrastructure?

```
git grep -P --show-function "\b(incrError|Abort)\b"
```

has some more matches.

Code:
```
+       sprintf msg, "Fatal: %s", message
```

This might break on <= IP7 due to sprintf limitations. `+` is better here as you don't know how long message is.

```
+       if(ParamIsDefault(abort) || abort)
```

Our default approach is to handle optional arguments after declaring variables and then later on use the result. You are also not forcing `abort` to a boolean value (0/1).

So something like `abort = ParamIsDefault(abort) ? 0 : !!abort` or with `if` in multiple lines.

49722281 (Abort test run if RTE ocurred in UTF, 2022-10-30)

That looks good!

Garados007:
I have changed the reporting functions. Now there are two function: `ReportError` and `ReportErrorAndAbort`. Also the whole code base is changed to use one of the two functions instead calling `incrError` and aborting themself. There is a single point left which is not changed. And that is `ReportResults` in `unit-testing-basics.ipf`.

As the result the function `abortNow()` is no longer really used. Only one example and `ReportResults` is calling it. But I cannot remove it because it is marked as public.

t-b:
>As the result the function abortNow() is no longer really used. Only one example and ReportResults is calling it. But I cannot remove it because it is marked as public.
 
Although it is public we don't advertise it as being part of our official API. Our public API is everything which is covered by doyxgen. We use the HIDDEN_SYMBOL `@cond/@endcond` approach to hide everything else from doxygen.

t-b:
Review:

- [ ] Nit pick: Please use spaces for alignment, we only indent with tabs. Yes the existing does it also wrong...
- [ ] ReportErrorAndAbort: Don't use `Abort message` as that pop ups a dialog which breaks non-interactive use. Just `Abort`  is fine.
- [ ] The use of `incrError()` in UTFBackgroundMonitor can also be replaced with ReportError I guess.
- [ ] There are still some `Abort`s in place, see `git grep --show-function -C3 "^[[:space:]]*Abort\b" procedures/`. I think all except the last one can use ReportErrorAndAbort. The last should actually just use `return 0` IMHO.
- [ ] Now we are getting close to feature creep, but there are a couple of places where `UTF_PrintStatusMessage` should be replaced with ReportError/ReportErrorAndAbort. Some can be found with ` git grep -A1 "sprintf msg, \"Error"`
- [ ] Please rebase against main to also nicify the changes from b608fc6f (Tracing: Handle TUFXOP erros in Z_ function, 2022-11-02).
- [ ] Can you add some developer documentation to EvaluateResults so what we know when to use this one and when ReportError/ReportErrorAndAbort.

Thanks!

t-b (changes requested):
See comment.

Garados007:
> Please rebase against main to also nicify the changes from https://github.com/byte-physics/igor-unit-testing-framework/commit/b608fc6f743b0f8bd4046a14174110dfd8d67afa (Tracing: Handle TUFXOP erros in Z_ function, 2022-11-02).
Thanks for pointing out this problem.

Garados007:
This PR requires PR #306 to be merged first.

t-b:
We are getting close :)

Questions:
- ReportError has a fatal parameter. The description says something about "prevent further testcase execution". But we are only adapting the message.
- There are some cases of setAbortFlag after "ReportError(... fatal = 1)".

Could we maybe fix the two issues in one go and always call `setAbortFlag()` when `fatal = 1`? Or does that not make sense?

Garados007:
> ReportError has a fatal parameter. The description says something about "prevent further testcase execution". But we are only adapting the message.

I have seen that some messages had the prefix "Fatal:" and handled some abortion routine. I have collected these together in `ReportErrorAndAbort`. Then I have seen some places with a "Fatal:" but not immediate abortion. So I have just marked these messages as "Fatal" and didn't abort anything to do not break with current behavior. Right now the parameter `fatal=1` is nothing more than marking the message as "Fatal".

I have found 3 places with the direct use of `fatal=1` (no call to `ReportErrorAndAbort`):

- https://github.com/byte-physics/igor-unit-testing-framework/blob/5d6d63d0d28cb510bbaecc9acb22ae8a1c54bf89/procedures/unit-testing-basics.ipf#L2368-L2373
- https://github.com/byte-physics/igor-unit-testing-framework/blob/5d6d63d0d28cb510bbaecc9acb22ae8a1c54bf89/procedures/unit-testing-basics.ipf#L2379-L2384
- https://github.com/byte-physics/igor-unit-testing-framework/blob/5d6d63d0d28cb510bbaecc9acb22ae8a1c54bf89/procedures/unit-testing-basics.ipf#L3664-L3670

I can refactor these places to use `ReportErrorAndAbort` with setting the abort flag but without calling `Abort` at the end. With this `ReportError` is no longer in need for the `fatal` parameter.

> - There are some cases of setAbortFlag after "ReportError(... fatal = 1)".
> 
> Could we maybe fix the two issues in one go and always call setAbortFlag() when fatal = 1? Or does that not make sense?

With the previous suggestion of removing the the option `fatal` and calling `ReportErrorAndAbort` this will be solved in one go.

Garados007:
What do you think about it?

t-b:
@Garados007 Sounds good!

t-b:
Would it make sense to have one function instead of `ReportError`/`ReportErrorAndAbort`? I don't know so I'm asking you.

Garados007:
@t-b Yes, it is possible but I don't think it's the right way. If I want to combine both function I have to add more parameter to the function which increases the complexity and reduce the readability - I have to look for each call at the end of the line to know what the call is actually doing. For some messages this can be troublesome if the message is placed inline and I have to scroll for approx. 200 columns (not everyone has a 4k monitor).

Having distinct methods is better:
1. I can look at the method call to know what is the intent of the call without scrolling to the end of the line.
2. It is easier to categorize the kind of messages. I have to only search for the method name to find similar calls and I don't have to use complex regular expressions.
3. Clean methods with less complex parameter. With this I can define the default behavior for the scenarios and use parameters only as an exception to this. This also allows to write less and clean code to achieve the same result.

t-b:
@Garados007 Okay, let's keep both functions.

t-b (approved):

t-b:
@Garados007 Thanks for your patience and endurance!

PR #297, opened 2022-11-01
Feature: Check for free wave leaks in test cases

### LINKED ISSUE #200: Investigate free wave leak tracking
- IP9 only
- Add optional parameter to RunTest, defaulting to 0 for IP9, when set for IP<9 then error
- Use `WaveTracking`
- Wave leaks are a CHECK error
- Add to Test Case Begin / End use incrError() on fail
- add new function tag to disable this check for a specific test case (when parameter was 1)

Garados007:
It is possible to write igor code that won't release free waves after they are no longer used. This can conclude to big wave leaks and therefore useless memory consumption.

With this commit the ability to check for these free wave leaks is introduced. To enable this it is only required to set `waveTrackingMode` in `RunTest` to `UTF_WAVE_TRACKING_FREE`, `UTF_WAVE_TRACKING_LOCAL` or `UTF_WAVE_TRACKING_ALL`. This will track the number of created and released waves during a test case run. If there is a difference this will produce a `CHECK` error.

If the user wants to opt-out a single test case from this wave tracking the test case as to be tagged with `UTF_NO_WAVE_TRACKING`.

Close #200.

## Dependency

- [x] PR #306: CI Bug
- [x] PR #310: Use new TestCaseFail methods
- [ ] Results in MIES
- [ ] WaveMetrics Bug Reports regarding Wave Tracking


Garados007:
I have resolved the merge conflict.

t-b:
@MichaelHuth I've assigned you instead due to workload issues on my side.

procedures/unit-testing-constants.ipf
```diff
@@ -137,3 +138,12 @@ StrConstant UTF_KEY_REGEXP               = "REGEXP"
 /// @{
 Constant UTF_MAX_PROC_LINES = 10000
 /// @}
+
+/// @name Wave tracking modes. These are flags that can be combined
```
MichaelHuth, line 142: "These are flags that can be combined" that is not fully correct, as you would have then only 1 and 2 as constant. What you mean is that the tracking modes are a bit pattern.
I personally tend to use for bits the hex notation through
0x01 and 0x02 etc. as a kind of visual differentiation to enums.

MichaelHuth, line 142: "These are flags that can be combined" that is not fully correct, as you would have then only 1 and 2 as constant. What you mean is that the tracking modes are a bit pattern.
I personally tend to use for bits the hex notation through
0x01 and 0x02 etc. as a kind of visual differentiation to enums.

MichaelHuth, line 142: "These are flags that can be combined" that is not fully correct, as you would have then only 1 and 2 as constant. What you mean is that the tracking modes are a bit pattern.
I personally tend to use for bits the hex notation through
0x01 and 0x02 etc. as a kind of visual differentiation to enums.

procedures/unit-testing-basics.ipf
```diff
@@ -3414,13 +3486,21 @@ End
 ///                         "IUTF_Test.log". If disabled the file names will always contain the name of the procedure file and a
 ///                         timestamp.
 ///
+/// @param   waveTrackingMode (optional) default disabled, enabled when set to something different to 0: @n
```
MichaelHuth, line 3489: "to a value different than 0"

procedures/unit-testing-basics.ipf
```diff
@@ -2245,6 +2286,33 @@ static Function AfterTestCase(name)
 
 	DFREF dfr = GetPackageFolder()
 	NVAR/SDFR=dfr assert_count
+	NVAR/SDFR=dfr waveTrackingMode
+
+#if IgorVersion() >= 9.0
+	if(NVAR_Exists(waveTrackingMode))
+		if((waveTrackingMode & UTF_WAVE_TRACKING_LOCAL) == UTF_WAVE_TRACKING_LOCAL)
+			WaveTracking/LOCL count
+			if(V_Flag == 1)
```
MichaelHuth, line 2295: Introduce a static Constant for the WaveTracking mode check.
What happens if the mode here is not 1? (i.e. the use code hampered with the wavetracking)
Currently you ignore that silently.
I think this should give an error as well, as the test was run with the purpose to check that and if the UTF can't do that because the test case is incompatible it should fail as well.

- Silently ignoring is not an option

procedures/unit-testing-basics.ipf
```diff
@@ -2245,6 +2286,33 @@ static Function AfterTestCase(name)
...
+				sprintf msg, "Local wave leak detected (leaked waves: %d) in \"%s\"", V_numWaves, name
```
MichaelHuth, line 2296: The sprintf msg generation can be moved into the following if.

procedures/unit-testing-basics.ipf
```diff
@@ -2245,6 +2286,33 @@ static Function AfterTestCase(name)
...
+				if(V_numWaves)
+					InternalHookFail(msg)
+				endif
+				WaveTracking/LOCL stop
+			endif
+		endif
+
+		if((waveTrackingMode & UTF_WAVE_TRACKING_FREE) == UTF_WAVE_TRACKING_FREE)
+			WaveTracking/FREE count
+			if(V_Flag == 1)
```
MichaelHuth, line 2306: see comment above

procedures/unit-testing-basics.ipf
```diff
@@ -2234,6 +2257,24 @@ End
 static Function BeforeTestCase(name)
 	string name
 
+	variable err
```
MichaelHuth, line 2260: Where is err used?

procedures/unit-testing-basics.ipf
```diff
@@ -2234,6 +2257,24 @@ End
...
+
+	DFREF dfr = GetPackageFolder()
```
MichaelHuth, line 2262: This block can be moved inside the #if

procedures/unit-testing-basics.ipf
```diff
@@ -2234,6 +2257,24 @@ End
...
+	NVAR/SDFR=dfr waveTrackingMode
```
MichaelHuth, line 2263: Use NVAR/Z if the global can potentially not exist.

procedures/unit-testing-basics.ipf
```diff
@@ -667,6 +667,29 @@ static Function UTF_ToSystemErrorStream(message)
 	systemErr += message + "\r"
 End
 
+/// Reports an error. This is used for internal UTF Hooks.
```
MichaelHuth, line 670: The description should state a bit clearer why this special function must be used from hooks to report errors.
Also does it work from all UTF hooks?

MichaelHuth (changes requested):
There is a conceptional point I do not understand:
You transfer the RunTest argument to the s. structure and store it as well in a global variable.
Why do you need to keep it also in the s. structure?

I think it would be simpler to transfer the argument that is in a local variable to the global at the start of the test run as you don't need the state saving through the s. structure.

t-b:
Before this is merged I would like to start a CI run with MIES to see if it is working as intended. I can think of data generators being a problem or user hooks.

procedures/unit-testing-basics.ipf
```diff
@@ -2245,6 +2289,43 @@ static Function AfterTestCase(name)
...
+		if((waveTrackingMode & UTF_WAVE_TRACKING_LOCAL) == UTF_WAVE_TRACKING_LOCAL)
+			WaveTracking/LOCL count
+			if(V_Flag == WAVE_TRACKING_COUNT_MODE)
+				if(V_numWaves)
+					sprintf msg, "Local wave leak detected (leaked waves: %d) in \"%s\"", V_numWaves, name
+					TestCaseFail(msg)
+				endif
+				WaveTracking/LOCL stop
+			elseif(V_Flag != WAVE_TRACKING_INACTIVE_MODE)
+				// do nothing for WAVE_TRACKING_INACTIVE_MODE.
```
MichaelHuth, line 2305: This creates some unintuitive situation. If you add on purpose to TestProduceWaveLeaks
`WaveTracking/FREE tracker` then the UTF fails for this TC, but also the next TC TestWaveLeaksIgnored fails.
Though, how can this happen when it is marked with `// UTF_NO_WAVE_TRACKING` ?

So if waveTrackingMode is enabled you have to either stop any user induced WaveTracking after the TC or
you stop it in BeforeTestCase even if the tag is set (when waveTrackingMode is enabled).

internal_dev/TracingZError.ipf
```diff
@@ -1,4 +1,4 @@
-﻿#pragma TextEncoding = "UTF-8"
+#pragma TextEncoding = "UTF-8"
```
MichaelHuth, line 1: Is this relevant for this PR?

MichaelHuth (changes requested):
There is still a small issue with the UTF controlling the WaveTracking.

internal_dev/TracingZError.ipf
Garados007: It is relevant to fix the current main branch to let this PR merge.

Garados007:
This PR requires PR #306 to be merged first.

MichaelHuth (approved):
Looks good.

Garados007:
I was forced to do a rebase to let the CI pipeline rerun. You need to approve this PR again.

t-b:
Michael and me played around with this PR and MIES. And found some bugs in IP. I've reported them as #4280/#4281. Once these are fixed I'll resume testing.

MichaelHuth:
@Garados007 Resulting testing with Thomas, I have moved the wavetrack evaluation to a function that is called after the user hook code. This is required as we need in tests the user hook to cleanup created waves to reach the same state as on test case entry.

t-b:
@Garados007 Can we have a new PR with only the introduction of `TestCaseFail`? And also maybe reuse the new ReportError/ReportErrorAndAbort functions? And avoid code duplication with printFailinfo? And also use TestCaseFail in other cases (I've found at least AfterTestCase). Thx.

t-b:
Note to self: Reported another issue to WM regarding correct tracking.

Garados007:
> Can we have a new PR with only the introduction of TestCaseFail?

See PR #310. The commit is still included in this PR (for compatibility issues). I remove it after #310 is merged to main.

> And also maybe reuse the new ReportError/ReportErrorAndAbort functions?

Done.

> And avoid code duplication with printFailinfo? And also use TestCaseFail in other cases (I've found at least AfterTestCase).

Let's move this to PR #310.

Garados007:
I have included the new TestCaseFail method into this PR.

t-b:
After some more playing with the MIES Basic.pxp tests, I found https://gitlab.byte-physics.de:4001/internal/json-xop/-/issues/125. This is very likely reponsible for the vast majority of SF wave leaks.

The JSON-XOP testsuite does not wave any wave leaks though.

t-b (changes requested):
See comment

t-b:
Oh and I don't think you need to `return NaN` after `ReportErrorAndAbort`. At least in other places you also skip the return.

t-b (approved):
LGTM.

PR #308, opened 2022-11-08
AreVariablesClose: Prevent singularity

### LINKED ISSUE #216: AreVariablesClose does not check properly if values are zero
We calculate d1 and d2 in AreVariablesClose through division:
https://github.com/byte-physics/igor-unit-testing-framework/blob/989148e15e06e001a1d407465f9e148425f01f0d/procedures/unit-testing-assertion-checks.ipf#L127

If the variable to check is zero then d1 and/or d2 is NaN.
Thus, the following if check for strong fails if one is Nan and for weak it fails if both are NaN.

Garados007:
If AreVariablesClose is called with one of both arguments zero it will
trigger a singularity when executing "diff / abs(var)". This is an issue
in the used algorithm. The boost framework provided a solution for this
case as such as catching this special case and using a different
algorithm. [1]

With this fix AreVariablesClose can be called with any real values.

[1]: https://www.boost.org/doc/libs/1_77_0/libs/test/doc/html/boost_test/testing_tools/extended_comparison/floating_point/floating_points_comparison_theory.html

Close #216.


t-b:
The solution itself makes sense. I would prefer to see `IsVariableSmall` be used instead of open coding it. Placing that into AreVariablesClose also makes sense for testing.

Garados007:
I don't think that `IsVariableSmall` is the right approach. The problem in the normal algorithm is that at one point the reciprocal of the variable is computed and that is defined for all real numbers except zero. I can use `IsVariableSmall` but I have to use a very very small tolerance. More precisely the smallest positive real number that a floating point number can express. And then there is only one number which absolute is smaller than this tolerance and that is: zero. For all numbers larger or equal this tolerance this algorithm is working.

Because this algorithm is based on the implementation of the algorithm in Boost I have looked more deeply into how they are solving the problem:

https://github.com/boostorg/math/blob/b2538faaf9802af8e856c603b9001e33db826676/test/test_tr1.c#L32-L48

It's pretty much the same like the solution I have proposed in this PR.

t-b:
I meant to use

```diff
$git diff .
diff --git a/procedures/unit-testing-assertion-checks.ipf b/procedures/unit-testing-assertion-checks.ipf
index c5f4628..e62930f 100644
--- a/procedures/unit-testing-assertion-checks.ipf
+++ b/procedures/unit-testing-assertion-checks.ipf
@@ -136,7 +136,7 @@ static Function AreVariablesClose(var1, var2, tol, strong)
        variable strong
 
        if(var1 == 0 || var2 == 0)
-               return abs(var1 - var2) <= tol
+               return IsVariableSmall(var1 - var2, tol)
        endif
 
        strong = !!strong

```

as I think this is the same except `<` vs `<=`.

Garados007:
@t-b  I have added more of these tests and I think one thing is weird:

```igorpro
Ensure(!UTF_Checks#AreVariablesClose(0, 0, 0, WEAK_CLOSENESS))
Ensure(!UTF_Checks#AreVariablesClose(0, 0, 0, STRONG_CLOSENESS))
```

Normally I would expect that 0 and 0 are close. But to succeed the test this has to be false. The problem is that the tolerance is 0 and a smaller error than 0 is required which is impossible.

I think we have to change the tolerance check to `<=`.

t-b:
> I think we have to change the tolerance check to <=. (like in Boost)

Good point. Yes please do that. But do that in a separate commit with accompanying tests. In principle this is a breaking chagne, but one we actually want.

t-b:
You would change IsVariableSmall directly right?

t-b:
And let's make a new PR for that as well, so that we can streamline that.

Garados007:
> You would change IsVariableSmall directly right?

Yes. In `AreVariablesClose` there is no need to change. It use `<=` already.

> And let's make a new PR for that as well, so that we can streamline that.

PR #309

t-b (approved):

PR #310, opened 2022-11-10
Reporting: Add TestCaseFail

Garados007:
Introduce a new convenient method to let a testcase without the use of
assertions fail. This method is only intended to be called outside a
user testcase.

See: PR #297

## Dependency

- [x] PR #298: Add Output to summary

### DIFF
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -726,6 +726,37 @@ static Function UTF_ToSystemErrorStream(message)
 	systemErr += message + "\r"
 End
 
+/// Make a test case fail. This method is intended to use outside the user code
+/// of the test case as such as it won't look in the stack trace which assertion
+/// triggered the error.
+///
+/// This method is a short version of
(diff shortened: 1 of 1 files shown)


Garados007:
I changed the assignment back to me to add `AfterTestCase´ to other places.

Garados007:
**TODO:**

- [ ] Add output to summary (depends on PR #298)

Garados007:
> [Comment from @t-b in PR #297](https://github.com/byte-physics/igor-unit-testing-framework/pull/297#issuecomment-1310131724)
> And avoid code duplication with printFailinfo? And also use TestCaseFail in other cases (I've found at least AfterTestCase).

I have combined `PrintFailInfo` and `TestCaseFail` as one is the one triggered by an assertion and the other one is the more general one.

Garados007:
> Well we do output something to the history for leaks/missing assertions, so we could store that and output again at the end of the run or?

There is a new function `AddFailedSummaryInfo` which enables to add a line of text to the summary at the end of the test run. Missing assertion is using this with this PR. For wave leaks is PR #297 responsible.

t-b (approved):

Issue #311, opened 2022-11-14
Expected failure is broken


t-b:
Testcase:

```
/// UTF_EXPECTED_FAILURE
Function Test()

	FAIL()
End
```

gives with 4afcfe12 (Merge pull request #310 from byte-physics/mb/refactor/test-case-fail-reporting, 2022-11-14)

```
•runtest("procedure", testcase = "Test")
  Start of test "Unnamed"
  Entering test suite "Procedure"
  Entering test case "Test"
  Expected Failure: 0: is false. Assertion "FAIL()" failed in Test (Procedure, line 6)
  Leaving test case "Test"
  Failed with 1 errors
  Leaving test suite "Procedure"
  Test finished with 1 errors
  End of test "Unnamed"
```

but with caadcd6c (Merge pull request #267 from byte-physics/fix-too-late-initialization, 2022-10-04)

```
•runtest("procedure", testcase = "Test")
  Start of test "Unnamed"
  Entering test suite "Procedure"
  Entering test case "Test"
  Expected Failure: 0: is false. Assertion "FAIL()" failed in line 6, procedure "Procedure"
  Leaving test case "Test"
  Finished with no errors
  Leaving test suite "Procedure"
  Test finished with no errors
  End of test "Unnamed"
```

Note the "with 1 errors" vs "with no errors".

PR #313, opened 2022-11-14
FIX: error counter for test assertions

Garados007:
During some refactoring of the handling of internal error messages the
error counter got broken. It will increase the error always by one for
each error printing even it was an expected error.

This is now fixed as such as an option was introduced to select if the
error counter should be increased (normally used by UTF errors) or not
(normally the case for UTF assertions).

Close: #311.

### DIFF
--- a/procedures/unit-testing-basics.ipf
+++ b/procedures/unit-testing-basics.ipf
@@ -739,20 +739,23 @@ End
 ///                 summary.
 /// @param hideInSummary (optional, default disabled) If set to non zero it will hide this message
 ///                 in the summary at the end of the test run.
-static Function TestCaseFail(message, [summaryMsg, hideInSummary])
+/// @param  incrErrorCounter (optional, default enabled) Enabled if set to a value different to 0.
+///                 Increases the internal error counter.
+static Function TestCaseFail(message, [summaryMsg, hideInSummary, incrErrorCounter])
 	string message
 	string summaryMsg
-	variable hideInSummary
+	variable hideInSummary, incrErrorCounter
 
 	DFREF dfr = GetPackageFolder()
 	SVAR/SDFR=dfr type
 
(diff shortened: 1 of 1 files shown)


Garados007:
Small example using this fix:

```igorpro
/// UTF_EXPECTED_FAILURE
Function Test()

	FAIL()
End

Function Foo()

	FAIL()
End
```

```
  Start of test "Unnamed"
  Entering test suite "Proc0"
  Entering test case "Test"
  Expected Failure: 0: is false. Assertion "FAIL()" failed in Test (Proc0, line 11)
  Leaving test case "Test"
  Entering test case "Foo"
  0: is false. Assertion "FAIL()" failed in Foo (Proc0, line 16)
  Leaving test case "Foo"
  Failed with 1 errors
  Leaving test suite "Proc0"
  Test finished with 1 errors
    ▶ Assertion "FAIL()" failed in Foo (Proc0, line 16)
  End of test "Unnamed"
```

Garados007:
Without this fix:
```
  Start of test "Unnamed"
  Entering test suite "Proc0"
  Entering test case "Test"
  Expected Failure: 0: is false. Assertion "FAIL()" failed in Test (Proc0, line 11)
  Leaving test case "Test"
  Entering test case "Foo"
  0: is false. Assertion "FAIL()" failed in Foo (Proc0, line 16)
  Leaving test case "Foo"
  Failed with 3 errors
  Leaving test suite "Proc0"
  Test finished with 3 errors
    ▶ Assertion "FAIL()" failed in Foo (Proc0, line 16)
  End of test "Unnamed"
```

Garados007:
That was indeed to quick. I forgot to set the default of the parameter which caused the new issue. 🤦‍♂️

t-b (approved):

Issue #318, opened 2022-11-17
Error popup as systemErr global is missing


t-b:
Steps:
- Open Igor Pro
- Paste

```igorpro

#include "unit-testing"

Function mytest()

End
```

into the builtin procedure window.

Execute

```igorpro
runtest("Procedure", enableJU=1)
```

on the commandline.

Observe:
```
While executing SVAR, the following error occurred: SVAR reference to "systemErr" failed.
```

Enabling the debugger and debug on error helps here.

PR #323, opened 2022-11-21
Bug-Fix: Fix undefined global var references

Garados007:
In 9144af0 (Simplify the report of internal errors, 2022-11-03) the
output for internal errors has been simplified. In some cases is
UTF_ToSystemErrorStream called before it's global variable is
initialized which produce a RTE error.

Fix: #318


Garados007:
> The fix does fix the problem. But I'm wondering why this solution here is superior to initializing the globals early enough?

There is no real meaning behind this. The provided solution was just the first thing that came into my mind when I thought about this. I have created PR #324 which solves this bug using early initialization. You can accept one or another.

PR #324, opened 2022-11-23
Bug-Fix: Fix undefined global var references

Garados007:
In 9144af0 (Simplify the report of internal errors, 2022-11-03) the
output for internal errors has been simplified. In some cases is
UTF_ToSystemErrorStream called before it's global variable is
initialized which produce a RTE error.

The global variables dfr:messageBuffer, dfr:message, dfr:type and
dfr:systemErr are now as early as possible initialized to prevent access
errors later.

Fix: #318


Garados007:
This PR is solving #318 using [Thomas suggestion](https://github.com/byte-physics/igor-unit-testing-framework/pull/323#issuecomment-1323314436).

t-b:
Jeep. LGTM.

t-b (approved):

PR #343, opened 2022-12-14
Prefix TAP description with the test case name

Garados007:
This makes it easier to understand the TAP output.

## Dependency

- [x] PR #327


Garados007:

t-b (approved):

PR #348, opened 2022-12-15
Use UTF_PrintStatusMessage instead of printf

### LINKED ISSUE #329: AnalyzeTracingResult: Use proper print routine
This function uses plain printf/fprintf but we do have UTF_PrintStatusMessage for that. This function also ensures that the message is printed to stdout. Fix similiar cases as well.

Garados007:
Most calls to printf or fprintf are changed to UTF_Reporting#UTF_PrintStatusMessage. This function also ensures that the message is printed to stdout.

UTF_Reporting#UTF_PrintStatusMessage is now marked as threadsafe.

Close #329

## Dependency

- [x] PR #327

### DIFF
--- a/procedures/unit-testing-assertion-wrappers.ipf
+++ b/procedures/unit-testing-assertion-wrappers.ipf
@@ -224,8 +224,8 @@ static Function NEQ_STR_WRAPPER(str1, str2, flags, [case_sensitive])
 	endif
 
 	result = !UTF_Checks#AreStringsEqual(str1, str2, case_sensitive)
-	tmpStr1 = UTF_Utils#PrepareStringForOut(str1)
-	tmpStr2 = UTF_Utils#PrepareStringForOut(str2)
+	tmpStr1 = UTF_Utils#IUTF_PrepareStringForOut(str1)
+	tmpStr2 = UTF_Utils#IUTF_PrepareStringForOut(str2)
 	sprintf str, "\"%s\" != \"%s\" %s case", tmpStr1, tmpStr2, SelectString(case_sensitive, "not respecting", "respecting")
 
 	EvaluateResults(result, str, flags)
--- a/procedures/unit-testing-autorun.ipf
+++ b/procedures/unit-testing-autorun.ipf
@@ -125,19 +125,20 @@ End
 /// in the same folder as this experiment as timestamped file "run_*_*.log"
 Function SaveHistoryLog()
 
-	string historyLog
+	string historyLog, msg
 	historyLog = GetBaseFilename() + ".log"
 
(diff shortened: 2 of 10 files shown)


Garados007:
I cannot make `UTF_PrintStatusMessage` threadsafe because it will fail with IMTest in Igor 8. The reason is that the line

```igorpro
	tmpStr = UTF_Utils#PrepareStringForOut(msg, maxLen = IP8_PRINTF_STR_MAX_LENGTH - 2)
```

cannot executed because it contains the full function name (module name + function name). I don't know why this limitation exists and I cannot find any reference in the documentation.

Garados007:
I can copy the required code in the Igor 8 block to make it work. Making the required methods public or removing the threadsafe attribute does not seems to be a solution for me.

t-b:
Does `PrepareStringForOut` really have to be threadsafe? I know we default to that nowadays, but I can't see an obvious reason.

Garados007:
1. `PrepareStringForOut` is used by `UTF_PrintStatusMessage` (unit-testing-reporting.ipf)
2. `UTF_PrintStatusMessage` is used by `Z_` (unit-testing-tracing-tracer.ipf)

`Z_` is marked as threadsafe. And #329 requested that all places that use printf to output something should use `UTF_PrintStatusMessage`. To enable this I have to mark `UTF_PrintStatusMessage` threadsafe.

t-b:
> `Z_` is marked as threadsafe. And #329 requested that all places that use printf to output something should use `UTF_PrintStatusMessage`. To enable this I have to mark `UTF_PrintStatusMessage` threadsafe.

Got it, thanks for the explanation.

Garados007:

t-b:
Review:

1e09553f (Use UTF_PrintStatusMessage instead of printf, 2022-12-15)

Mostly Good.

The change

```diff
-static Function/S PrepareStringForOut(str, [maxLen])
+threadsafe static Function/S PrepareStringForOut(str, [maxLen])
```

looks a bit odd here.

cd9a3545 (Fix UTF_PrintStatusMessage to be threadsafe in Igor 8, 2022-12-15)

- [ ] This commit needs to come before.
- [ ] Why are we now using PrepareStringForOut also for IP >= 9?
- [ ] IP8_PRINTF_STR_MAX_LENGTH was changed without justification
- [ ] Is the code duplication here really better than making PrepareForStringOutput non-static?

t-b (changes requested):
See comment

Garados007:
I made PrepareForStringOutput public and therefore has most of this PR changed. Please review again @t-b @MichaelHuth .

t-b (approved):

PR #351, opened 2022-12-15
unit-testing-autorun.ipf: Add missing ModuleName pragma

t-b:
This allows to use code coverage determination to be used. Without it the UTF complains that AfterFileOpenHook is static and, without a module name, unaccessible.


MichaelHuth (approved):
ok

PR #352, opened 2022-12-20
GetWavePointer_Impl: Fix return memory address

t-b:
We don't want to prefix it with 08 but instead we want to have 08 as precision which must stand *after* the percent sign.

While at it we also leverage the power of # and capital X to get a capital 0X prefix.

Bug introduced in b7bb12a4 (AreWavesEqual creates descriptive message in WAVE_DATA mode, 2021-03-04).

### DIFF
--- a/procedures/unit-testing-utils.ipf
+++ b/procedures/unit-testing-utils.ipf
@@ -477,9 +477,9 @@ threadsafe static Function/S GetWavePointer_Impl(wv)
 	WAVE ref = refWave
 
 #if IgorVersion() >= 7.0
-	sprintf str, "0x08%x", ref[0]; err = GetRTError(1)
+	sprintf str, "%#08X", ref[0]; err = GetRTError(1)
 #else
-	sprintf str, "0x08%x", ref[0]
+	sprintf str, "%#08X", ref[0]
 #endif
 
 	return str


MichaelHuth (approved):
Yep, that is more like it should be.

Garados007:
I found an interesting bug but I don't know why the [pipeline](https://gitlab.byte-physics.de:4001/mirror/igor-unit-testing-framework/-/pipelines/7273) for [b3ca346](https://github.com/byte-physics/igor-unit-testing-framework/pull/352/commits/b3ca346c76a83f4fdf5ce7d6631cdbc3d4ae6d38) succeeded. If I check out this commit (the same goes for main) and run Various.ipf locally I get the following error:

```
  Entering test case "UTF_Various#TC_WaveName"
  0: is false. Assertion "CHECK(GrepString(str, "^_free_ \\(0x[0-9a-f]+\\)$"))" failed in TC_WaveName (Various.ipf, line 574)
  0: is false. Assertion "CHECK(GrepString(str, "^namedFreeWave \\(0x[0-9a-f]+\\)$"))" failed in TC_WaveName (Various.ipf, line 579)
  Leaving test case "UTF_Various#TC_WaveName"
```

If I add `INFO("name: \"%s\"", s1 = str)` right before the `CHECK` I get more information:

```
  Entering test case "UTF_Various#TC_WaveName"
  0: is false. Assertion "CHECK(GrepString(str, "^_free_ \\(0x[0-9a-f]+\\)$"))" failed in TC_WaveName (Various.ipf, line 575)
    ⓘ name: "_free_ (0XFD735D0)"
  0: is false. Assertion "CHECK(GrepString(str, "^namedFreeWave \\(0x[0-9a-f]+\\)$"))" failed in TC_WaveName (Various.ipf, line 581)
    ⓘ name: "namedFreeWave (0XFD73E30)"
  Leaving test case "UTF_Various#TC_WaveName"
```

It seems that the format of the address has been changed. Everything has now capital letters. I will address this issue in PR #355 but I am still puzzled why this was allowed to pass in CI.

A quick side note by myself: I think it is more difficult to read the pointer if the "x" is capital.

Garados007:
I have looked more in the results of the pipeline run. The test for `Various` didn't even run till the end. It just ended midway. These are the last lines in the log:

```
  Entering test case "UTF_Various#TC_MMD_Types:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0"
  Leaving test case "UTF_Various#TC_MMD_Types:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0"
  Entering test case "TC_MD_bck:first"
```

The test run is considered as succeeded because the failed JU file is never written to disk.

Garados007:
Found more hints of this bug. I have created #356 to tackle this.

t-b:
@Garados007 Thanks for noticing! Here Various.pxp finishes locally but does not pass (which is expected).

PR #372, opened 2023-01-05
Documentation: Add custom.css with more spacing between functions

t-b:


docu/sphinx/source/_static/custom.css
```diff
@@ -0,0 +1,3 @@
+dl.function {
+margin-bottom: 50px;
```
Garados007, line 2: I think this can be indented with two spaces.

Garados007, line 2: I don't like `px` units as it has issues with mobile screens, high resolution screens and so on. It is recommended to use `em` or `rem`. In this case `50px` would be `2,9412em` but you can just say `3em`.

docu/sphinx/source/_static/custom.css
```diff
@@ -0,0 +1,3 @@
+dl.function {
```
Garados007, line 1: Doing this in this way you have the first entry flushed to the headline and the last entry will leave a big gap to the footer. There are two solutions:

1. Split the margin into `margin-top` and `margin-bottom`.
2. Use `dl.function:not(:last-child)` as the selector.

I recommend the second one.

Garados007 (changes requested):
I have some improvements to your CSS code.

The links seems to working with the local sphinx build.

docu/sphinx/source/_static/custom.css
t-b: Thanks for the review. I've applied all suggestions! See https://docs-staging.byte-physics.de/igor-unit-testing-framework which I deployed by executing the deployment-staging job.

Garados007: Building this locally on my computer produces the correct page:

grafik

But when I look at the staging branch using your link I will still see the old code:

grafik

Garados007: The artifacts from `documentation` step are correct. The error must be in `deployment staging`.

Garados007 (approved):
The current changes are okay and ready to merge. ~There is only the issue left that docs-staging.byte-physics.de contains wrong code.~ *(Browser Cache ¯\\_(ツ)_/¯)*

docu/sphinx/source/_static/custom.css
Garados007: My Browser Cache tricked me. It's working now. We can use version specific links for assets to prevent this in the future but I don't know if we change the assets that often.

t-b: > We can use version specific links for assets to prevent this in the future but I don't know if we change the assets that often.

Normally we don't chang them often.

PR #377, opened 2023-01-11
Require UTF_EXPECTED_FAILURE to have errors

### LINKED ISSUE #289: Should UTF_EXPECTED_FAILURE require the test to fail?
```
#include "unit-testing"

// UTF_EXPECTED_FAILURE
Function Test()

	PASS()
End
```

Generates 

```
•runtest("Procedure")
  Start of test "Unnamed"
  Entering test case "Test"
  Leaving test case "Test"
  Finished with no errors
  Test finished with no errors
  End of test "Unnamed"
```

which I find a bit suprising. Should we require that an expected failure test fails?

Garados007:
Each test case that in now marked with UTF_EXPECTED_FAILURE has to create a failure or error at least once.

Close #289

## Dependency

- [x] PR #327 
- [x] PR #381


Garados007:

t-b:
Review:

- Why do we need to set `wvTestCase[%CURRENT][%NUM_ASSERT_ERROR] = "0"` in `AfterTestCase`? This was not done before?
- Would it make sense to remove the logError parameter of TestCaseFail as you removed the last caller?

t-b (changes requested):
See comment

Garados007:
> Why do we need to set wvTestCase[%CURRENT][%NUM_ASSERT_ERROR] = "0" in AfterTestCase? This was not done before?

No. The updated implementation count errors as normal to allow the check 5 lines above this statement. After that we have to clear the error counter as these errors are considered as intended and an existing error counter is propagated upwards which let the whole test run fail.

> Would it make sense to remove the logError parameter of TestCaseFail as you removed the last caller?

Thanks for the reminder. I have to update the documentation as well.

I will pause further reviews until PR #381 is merged. PR #381 has changed the location of some hooks which results in heavy rebases.

Garados007:
@t-b @MichaelHuth Ready for next review round.

t-b (approved):
LGTM

Issue #54, opened 2019-03-20
JUNIT: Revise error count for failing user test hooks


t-b:
Running `AbortingUserHooks.pxp` gives the following JUNIT file

```xml
<?xml version="1.0" encoding="utf-8"?>
<testsuites>
        <testsuite package="aborting-user-hooks.ipf" id="0" name="aborting-user-hooks.ipf" timestamp="2019-03-20T11:39:16Z+01:00" hostname="
THOMAS-WIN7-X64" tests="1" failures="0" errors="1" skipped="0" disabled="0" time="0.027">
                <properties>
                        <property name="User" value="thomas"/>
IGORFILEVERSION:8.0.3.2"/>roperty name="System" value="OS:Windows 7 Professional Service Pack 1;
                        <property name="Experiment" value="AbortingUserHooks"/>
                        <property name="UTFversion" value="1.08"/>
SCREEN3:DEPTH=32,RECT=2400,0,3600,1920"/>gorInfo" value="IGORVERS:8.03;
                </properties>
                <testcase name="Dostuff in aborting-user-hooks.ipf (7)" classname="Dostuff" time="0.007">
                        <error message="Encountered &quot;Abort&quot; in user hook &quot;TEST_CASE_END_OVERRIDE&quot;, procedure file &quot;
" type="ABORT"></error>&quot;
</system-out>case &quot;TEST_CASE_END_OVERRIDE&quot;T_CASE_END_OVERRIDE&quot;, procedure file &quot;aborting-user-hooks.ipf&quot;t;
                <system-err>Encountered &quot;Abort&quot; in user hook &quot;TEST_CASE_END_OVERRIDE&quot;, procedure file &quot;aborting-use
</system-err>uot;
                </testcase>
</system-out>case &quot;TEST_CASE_END_OVERRIDE&quot;T_CASE_END_OVERRIDE&quot;, procedure file &quot;aborting-user-hooks.ipf&quot;t;
                <system-err>Encountered &quot;Abort&quot; in user hook &quot;TEST_CASE_END_OVERRIDE&quot;, procedure file &quot;aborting-use
</system-err>uot;
        </testsuite>
</testsuites>
```

but we actually had 6 errors as all user hooks aborted.
Should/Can we cramp all errors into a testsuite? Or should we introduce errors/failures attributes to the toplevel `testsuites` element.

According to https://llg.cubic.org/docs/junit/ we could add these new attributes.

PR #381, opened 2023-01-12
 Treat user hooks as test cases

### LINKED ISSUE #344: Refactor ExecuteHooks to have only one switch statement
The `ExecuteHooks` function in `unit-testing-basics.ipf` contains two switch statement which iterate over the same variable. The only difference is that the first one is error checked with a try-endtry-statement and the second is not. Refactoring to one switch statement reduces the complexity of this function and make it easier to understand.

During the refactoring it could be that two internal hook functions will be called one after another. It should be checked if these functions can be merged together. This will also reduce the complexity with our hook setup.

### LINKED ISSUE #359: Treat user hooks as test cases
See #353 

This allows to track assertion failures in test hooks correctly.

TEST_BEGIN/TEST_END: It's own test suite with one test cases each
TEST_SUITE_BEGIN/TEST_SUITE_END: The very first test case of each test suite and the very last one.
TEST_CASE_BEGIN/TEST_CASE_END: One testcase before and after the real test case.

Will only be added with user hooks present.

Garados007:
Assertions in User Hooks are now working again and with a clean output in JUnit or TAP. At the same time the code is a bit refactored to allow these modification and make it more cleaner and easier to maintain. During these steps the solution for Issue #344 is included as it helped to implement #359.

Close #344
Close #359


Garados007:

Garados007:
I found a small bug in the results output while checking #54. I have included the fix and updated the documentation for this.

I will add #54 to the close list of this PR as the main topic (number of user hook abort errors inside JUnit) is solved here.

t-b:
Review:

a748b2a9 (Refactor test reporting results controls, 2023-01-12)

- SetupTestRun: Calling a function from wave getter to fill in the details is fine. But please just pass the wave reference. The current code can easily result in an stack recursion error if something is buggy.

99f2fdb1 (Rename unit-testing-hooks to unit-testing-hooks-proto, 2023-01-12)
5017e0de (Refactor Hooks into own file, 2023-01-12)

Looks all sensible.

729a8c38 (Hooks: Refactor ExecuteHooks to be cleaner, 2023-01-12)

> +/// @returns 1 if user hook succeed without an error and 0 if an error happened

- Please switch that. We prefer the unix logic of return 0 being successful and everything else an error. This also allows for more error codes. And also then initialize the `result` variable, which unfortunately is needed here, to 1. The default for a result variable should always be the error state.

From ExecuteHooks:

```igorpro
			if(ExecuteUserHook(name, userHook, procWin, HOOK_LEVEL_TEST_CASE))
				AfterTestCaseUserHook(name, param)
			endif

```

Two questions:
- Why do we do that even for skipped test cases?
- Why do we call `AfterTestCaseUserHook` only on sucessfull hook invocation? We seem to never do that in other cases?

- Can we have a test which verifies that the current behaviour for skipped test cases and the test case hooks stays constant?

7160539b (Treat user hooks as test cases for the results, 2023-01-13)

- For determining if a funcref is initialized you can use `FuncRefInfo` instead of parsing the name.

ec24ad06 (Refactor: Remove logTestCase parameter from EvaluateRTE, 2023-01-16)

Good call!

> I will add https://github.com/byte-physics/igor-unit-testing-framework/issues/54 to the close list of this PR as the main topic (number of user hook abort errors inside JUnit) is solved here.

I fail to see who this fixes #54. Please explain that in a commit message.

Garados007:
> Why do we do that even for skipped test cases?

That's an interesting bug. While performing all refactor processes I tried to keep old behavior as much as possible so I haven't changed this logic. This is a bug that is also in main.

I'll create an issue for this.

> Why do we call `AfterTestCaseUserHook` only on sucessfull hook invocation? We seem to never do that in other cases?

Same as above. In main `AfterTestCaseUserHook` is skipped when the user hook exists with an abort or RTE.

I'll create an issue for this.

> Can we have a test which verifies that the current behaviour for skipped test cases and the test case hooks stays constant?

I need more clarification. The current behavior with skipped tests is that `TEST_CASE_BEGIN_OVERRIDE` is not executed but `TEST_CASE_END_OVERRIDE` is. You can use this as simple test:

```igorpro
#include "unit-testing"

Function TEST_CASE_BEGIN_OVERRIDE(string name)
	printf "hook begin: %s\n", name
End

Function TEST_CASE_END_OVERRIDE(string name)
	printf "hook end: %s\n", name
End

// UTF_SKIP
Function TestCase()
	print "test case"
	PASS()
End
```
```igorpro
// cac6bdf (Merge pull request #348 from byte-physics/mb/fix/329-print-routines, 2023-01-16)
•RunTest("Procedure")
  Start of test "Unnamed"
  hook end: TestCase
  Finished with no errors
  Test finished with no errors
  End of test "Unnamed"
// dea03ce (Refactor: Remove logTestCase parameter from EvaluateRTE, 2023-01-16)
•RunTest("Procedure")
  Start of test "Unnamed"
  hook end: TestCase
  Finished with no errors
  Test finished with no errors
  End of test "Unnamed"
```

Or did you have something else in mind?

> I fail to see who this fixes https://github.com/byte-physics/igor-unit-testing-framework/issues/54. Please explain that in a commit message.

Fixed.

Garados007:
> For determining if a funcref is initialized you can use FuncRefInfo instead of parsing the name.

Good idea. I tried it but doesn't work. It's for both cases the same output:

```
NAME:TEST_CASE_END_OVERRIDE;ISPROTO:0;ISXFUNC:0;
NAME:TEST_SUITE_END;ISPROTO:0;ISXFUNC:0;
```

I keep the name parsing.

t-b:
> Or did you have something else in mind?

I think the test belongs then to PR for #384. Thanks for for managing the issues.

> > For determining if a funcref is initialized you can use FuncRefInfo instead of parsing the name.

> Good idea. I tried it but doesn't work. It's for both cases the same output:

> ```
> NAME:TEST_CASE_END_OVERRIDE;ISPROTO:0;ISXFUNC:0;
> NAME:TEST_SUITE_END;ISPROTO:0;ISXFUNC:0;
> ```

Got it. I guess SetDefaultHooks is reponsible for that. Let's keep your approach.

Review round 2:

ExecuteUserHook: The documentation says

> The return is still 1 as if no error happened.

but that is wrong as 0 is now success.

Rest looks good!

t-b (changes requested):
See comment

Garados007:
Fixed.

t-b (approved):
LGTM.

PR #382, opened 2023-01-12
Rewrite of Readme.md

MichaelHuth:
This is the rewrite of the Readme.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1,59 +1,81 @@
-![](https://img.shields.io/github/contributors-anon/byte-physics/igor-unit-testing-framework?style=plastic)
-![](https://img.shields.io/github/license/byte-physics/igor-unit-testing-framework?style=plastic)
+# Igor Pro Universal Testing Framework
 
-This project aims at providing a complete set of tools for programmers for writing and maintaining unit tests.
+[![Contributors](https://img.shields.io/github/contributors-anon/byte-physics/igor-unit-testing-framework?style=plastic)](https://github.com/byte-physics/igor-unit-testing-framework/graphs/contributors)
+[![GitHub release (latest by date)](https://img.shields.io/github/v/release/byte-physics/igor-unit-testing-framework?style=plastic)](https://github.com/byte-physics/igor-unit-testing-framework/releases)
+[![License](https://img.shields.io/github/license/byte-physics/igor-unit-testing-framework?style=plastic)](https://github.com/byte-physics/igor-unit-testing-framework/blob/main/License.txt)
 
-# Features:
-* Runs on Windows and MacOSX
-* Completely documented and with lots of examples
-* Includes more than ten test assertions coming in three versions
-* Allows for the arbitrary grouping of test cases
-* Easily expandable and adaptable
-* Support for executing test suites from the command line
+The **Igor Pro Universal Testing Framework** (IUTF) is a versatile tool to write and run tests for [Igor Pro](https://www.wavemetrics.com/products/igorpro) code.
+The results of the tests is outputted in standardized formats that can be parsed by common
+[CI](https://en.wikipedia.org/wiki/Continuous_integration) frameworks. Test runs can also be fully automated as part of a CI/CD workflow.
+For simpler usage scenarios a summary of the test run is shown in the history after it finished.
 
-Planned features (open for collaboration):
-* Threadsafe test assertions
-* Igor Pro help file documentation
+## What can IUTF be used for?
 
-# Documentation
+The IUTF offers great flexibility for writing test cases. It is supported to run the same test case with multiple inputs as
+well as Igor Pro specific features such as independent modules or background activities.
 
-The documentation can be found
-[here](https://docs.byte-physics.de/igor-unit-testing-framework/). It contains
-a [guided
-tour](https://docs.byte-physics.de/igor-unit-testing-framework/guided-tour.html)
-and an [introduction to the basic
-structure](https://docs.byte-physics.de/igor-unit-testing-framework/basic.html).
+The IUTF is applied as a reliable tool in larger Igor Pro projects such as [MIES](https://github.com/AllenInstitute/MIES).
 
-# Requirements
+```igorpro
+#pragma TextEncoding="UTF-8"
+#pragma rtGlobals=3
 
-Igor Pro version 6.2.0 or later
+#include "unit-testing"
 
-# Installation
+// Simple demo example that tests the abs() function
+Function TestAbsolute()
 
-1. Install Igor
-2. Start Igor, this will create a folder called WaveMetrics in Documents
-3. Extract the zip file into the a folder, e. g. `thomas/unitTestingFramework` somewhere on your disc
-4. Create a link from `unitTestingFramework/procedures` to `Documents\WaveMetrics\Igor Pro 6 User Files\User Procedures`
-5. Have a look at the manual or the example experiments
-
-# Building the documentation
-
-## Requirements
-The build process is fully automated using [docker containers](https://www.docker.com/). You will need
-- [docker](https://www.docker.com/get-docker)
-- [git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
-
-## Building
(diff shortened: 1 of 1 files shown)


t-b:
Upps. I thouht I was assigned. Sorry for that. I've added some minor fixes and some more links.

t-b:
And thanks of course for the PR!

Garados007:
Do you really want to close this PR?

t-b:
Nope, of course not. I'll make myself a coffee ;)

Readme.md
```diff
@@ -1,59 +1,60 @@
...
 
-This will create a docker container with all the required dependencies and output the manual as pdf to `docu/manual.pdf` and html to the `docu/sphinx/html` subdirectory.
-The documentation is built using [doxygen](http://www.doxygen.org/), a [home-built awk script](https://github.com/byte-physics/doxygen-filter-ipf/), [breathe](https://github.com/michaeljones/breathe) and [sphinx](http://www.sphinx-doc.org).
+# Installation
 
-[The current documentation can be found on our website.](https://docs.byte-physics.de/igor-unit-testing-framework/)
+1. Extract the zip file into a folder, e. g. `alice/IgorTestingFramework`
+2. Open the Igor Pro user folder by selecting `Help/Show Igor Pro User Files` from the menu in Igor Pro
+3. Create a link from `alice/IgorTestingFramework/procedures` in the `User Procedures` subfolder
+4. Have a look at the [documentation](https://docs.byte-physics.de/igor-unit-testing-framework) or the example experiments
```
Garados007, line 60: This sounds a bit weird. Can you write something like: "Installation finished. You are ready to use IUTF in Igor Pro. Have a look..."?

Readme.md
```diff
@@ -1,59 +1,60 @@
-![](https://img.shields.io/github/contributors-anon/byte-physics/igor-unit-testing-framework?style=plastic)
-![](https://img.shields.io/github/license/byte-physics/igor-unit-testing-framework?style=plastic)
+[![Contributors](https://img.shields.io/github/contributors-anon/byte-physics/igor-unit-testing-framework?style=plastic)](https://github.com/byte-physics/igor-unit-testing-framework/graphs/contributors)
+[![GitHub release (latest by date)](https://img.shields.io/github/v/release/byte-physics/igor-unit-testing-framework?style=plastic)](https://github.com/byte-physics/igor-unit-testing-framework/releases)
+[![License](https://img.shields.io/github/license/byte-physics/igor-unit-testing-framework?style=plastic)](https://github.com/byte-physics/igor-unit-testing-framework/blob/main/License.txt)
 
-This project aims at providing a complete set of tools for programmers for writing and maintaining unit tests.
+The Igor Pro Universal Testing Framework (IUTF) is a versatile tool to write and run tests for [Igor Pro](https://www.wavemetrics.com/products/igorpro) code.
```
Garados007, line 5: Can you highlight the name **Igor Pro Universal Testing Framework**? You can use `**foo**` to achieve that.

Readme.md
```diff
@@ -1,59 +1,60 @@
-![](https://img.shields.io/github/contributors-anon/byte-physics/igor-unit-testing-framework?style=plastic)
-![](https://img.shields.io/github/license/byte-physics/igor-unit-testing-framework?style=plastic)
+[![Contributors](https://img.shields.io/github/contributors-anon/byte-physics/igor-unit-testing-framework?style=plastic)](https://github.com/byte-physics/igor-unit-testing-framework/graphs/contributors)
```
Garados007, line 1: Can you add the project title has a headline? You can do this with:

```markdown
# Igor Pro Universal Testing Framework
```

After that you can list the tags.

Readme.md
```diff
@@ -1,59 +1,60 @@
...
+[![GitHub release (latest by date)](https://img.shields.io/github/v/release/byte-physics/igor-unit-testing-framework?style=plastic)](https://github.com/byte-physics/igor-unit-testing-framework/releases)
+[![License](https://img.shields.io/github/license/byte-physics/igor-unit-testing-framework?style=plastic)](https://github.com/byte-physics/igor-unit-testing-framework/blob/main/License.txt)
 
-This project aims at providing a complete set of tools for programmers for writing and maintaining unit tests.
+The Igor Pro Universal Testing Framework (IUTF) is a versatile tool to write and run tests for [Igor Pro](https://www.wavemetrics.com/products/igorpro) code.
+The results of the tests is outputted in standardized formats that can be parsed by common
+[CI](https://en.wikipedia.org/wiki/Continuous_integration) frameworks. Test runs can also be fully automated as part of a CI/CD workflow.
+For simpler usage scenarios a summary of the test run is shown in the history after it finished.
```
Garados007, line 9: Can we split the following part in a new section? The current one seems a bit long. The new section could be something like "What is IUTF?" or "What can IUTF be used for?"

Readme.md
```diff
@@ -1,59 +1,60 @@
...
-Run `make` from the documentation root directory `docu/`:
-```bash
-cd docu
-make
-```
+Igor Pro version 6.37 or newer on Windows and MacOSX
 
-This will create a docker container with all the required dependencies and output the manual as pdf to `docu/manual.pdf` and html to the `docu/sphinx/html` subdirectory.
-The documentation is built using [doxygen](http://www.doxygen.org/), a [home-built awk script](https://github.com/byte-physics/doxygen-filter-ipf/), [breathe](https://github.com/michaeljones/breathe) and [sphinx](http://www.sphinx-doc.org).
+# Installation
```
Garados007, line 56: I think we have to add a step before the current first one:

```markdown
1. Download the zip file attached to the newest [Release](https://github.com/byte-physics/igor-unit-testing-framework/releases). The file name is something like `UnitTestingFramework-v1.09.zip`.
```

Garados007 (changes requested):
It is also common to add the following sections:

- **License**. In this case we can say that most parts are BSD-3. @t-b can extend this section in his PR to reference the license file.
- **Versioning**. Are we using [Semantic Versioning](http://semver.org/)? I think so.
- **Useful Links**. We can list some good links that can help the user to use IUTF or Igor Pro.

Garados007:
I can also create a fix if its okay.

t-b:
> It is also common to add the following sections:
> 
>     * **License**. In this case we can say that most parts are BSD-3. @t-b can extend this section in his PR to reference the license file.

Yes.

> 
>     * **Versioning**. Are we using [Semantic Versioning](http://semver.org/)? I think so.

Yes.

> 
>     * **Useful Links**. We can list some good links that can help the user to use IUTF or Igor Pro.

No, please not. This is too vague of a topic.

t-b:
@MichaelHuth Can we drop the WARN from the example? Thanks.

Garados007:
I have added my ideas. Please take a look.

Garados007 (approved):

MichaelHuth:
Looks good to me.

Just out of curiosity, if we use semantic versioning shouldn't we be at 1.9 ?
Does the single digit leading zero for the minor version number has some intention?

Garados007:
> if we use semantic versioning shouldn't we be at 1.9 ?

Yes there are some issues with the current versioning system that we use. The only exceptions that we have right now:

- We include leading zeros in older releases. Newer releases won't have this issue. ([according to Rule 2](https://semver.org/spec/v2.0.0.html#spec-item-2))
    I don't know why we have a leading zero at all.
- We doesn't include the patch version number. ([also Rule 2](https://semver.org/spec/v2.0.0.html#spec-item-2))

So, you are right we are not really following the rules of Semantic versioning as we break with rule 2.

I don't know. Maybe we should describe our versioning system.

t-b:
How about we just say that we want to use semantic versioning in the future and link to an issue?

I did that just now. I also revised the example a bit. And added Max and me as co authors to the commit.

t-b (approved):

t-b:
Please approve if you think this is good. Otherwise we iterate more.

t-b:
@MichaelHuth As you can't approve, it is your PR, does it look good now?

PR #383, opened 2023-01-16
Function Tags

### LINKED ISSUE #328: Bug: Cannot set function tag "#TAPDescription"
If I use the following code:

```igorpro
// #TAPDescription: Checking if '*' is the Answer to the Ultimate Question of Life, the Universe, and Everything
static Function TAPCheckNumber_not_nan()

	CheckNumber(char2num("*"))
End
```

*Source: https://github.com/byte-physics/igor-unit-testing-framework/blob/main/internal_dev/IM_tap_example1a-uncaught-aborts.ipf#L26-L30*

I get the following error:

```
error: A dimension label can't contain any of the following: single-quote, double-quote, semicolon, colon, or any control characters.
```

This happens because [`GetFunctionTagWave()`](https://github.com/byte-physics/igor-unit-testing-framework/blob/main/procedures/unit-testing-utils.ipf#L182) is trying to define a dimension label with the name `#TAPDescription:`.

Similar is with the function tag `#TAPDirective:`.

### LINKED ISSUE #378: Invalid UTF_EXPECTED_FAILURE Tag in Igor 6 and 7
If I write the following test cases:

```igorpro
// UTF_EXPECTED_FAILURE
static Function TestCase1()
	PASS()
End

static Function TestCase2()
	// other stuff
End
```

Both test cases are marked as UTF_EXPECTED_FAILURE when I look into `root:Packages:UnitTesting:TestRunData`. This is only the case for Igor 6 and 7.

Garados007:
This PR contains the following changes:

- Function tag utility functions are now refactored into a single file. This make it easier to manage them as we don't have to look in multiple files (UTF_Basics and UTF_Utils).
    f632d98 (Refactor: Move function tags functions to own file, 2023-01-16)
- Very long function names get now a short version when using them as dimension label to store function tags.
    fc2f76a (Function Tags: Fix long function name bug, 2023-01-16)
- Function tags are now sanitized to fix new dimension label behavior in Igor 9.
    e3c9940 (Function Tags: Cleanup function tags, 2023-01-16)

Close #378
Close #328


Garados007:
@t-b @MichaelHuth Ready for Review.

t-b:
Review:

f632d98c (Refactor: Move function tags functions to own file, 2023-01-16)

Looks good.

fc2f76ac (Function Tags: Fix long function name bug, 2023-01-16)

The approach here with the quick and long path seems quite complicated. Why not just use a 1D textwave for storing the full function names?

e3c99406 (Function Tags: Cleanup function tags, 2023-01-16)

I think CleanupName could replace SanitizeName or?

> The representation of the function tag itself is kept to keep compability with older experiments.

We don't need that. In case it would your live easier.

Tests are missing for both fixes.

t-b (changes requested):
See comment

t-b:
Another comment:

In GetFunctionTagWave the line

```
tag_constants[][1] = SanitizeTagName(tag_constants[p][0])
```

will result in the cleaned up name being found. But I don't think this is what we want.

Garados007:
@t-b 

> The approach here with the quick and long path seems quite complicated. Why not just use a 1D textwave for storing the full function names?

Do you want to throw dimension labels away at all and use FindValue in one wave (text wave) to find the index which is at the same time the same index in the second wave (wave reference wave)? My fix tried to keep the old behavior as much as possible and used the part with the dimension labels to its limits.

> We don't need that. In case it would your live easier.

That's good. It is easier to change the tag to a clean one in the first place than the part with the cleanup. So I will change this completely.

t-b:
> Do you want to throw dimension labels away at all and use FindValue in one wave (text wave) to find the index which is at the same time the same index in the second wave (wave reference wave)? My fix tried to keep the old behavior as much as possible and used the part with the dimension labels to its limits.

Yes that would be my favourite.

Garados007:
@t-b @MichaelHuth Ready for next review round.

t-b (approved):
LGTM. Nice!

PR #389, opened 2023-01-20
Introduce data generators as virtual test cases

### LINKED ISSUE #353: Test assertion in data generators
Found while playing around with #327:

```igorpro
#include "unit-testing"

Function/WAVE Gen()

	CHECK(0)
	
	Make/FREE/N=1 wv
	return wv
End

/// UTF_TD_GENERATOR Gen
Function MyTest([variable var])

	PASS()
End
```

gives

```
•runtest("procedure")
  0: is false. Assertion "CHECK(0)" failed in Gen (Procedure, line 8)
  Start of test "Unnamed"
  Entering test case "MyTest:0"
  Leaving test case "MyTest:0"
  Finished with no errors
  Test finished with no errors
  End of test "Unnamed"
```

which is unexpected as the failed assertion is not taken into account into the end result. Now knowing the UTF code a bit this is not too suprising. And I actually think it is nice to be able to use assertions in data generators. Or what do you think?

Same result in main, f69bf5d1 (Merge pull request #352 from byte-physics/bugfix/wave-pointer-printing, 2022-12-20), or mb/refactor/316-data-storage 7f659630 (Docs: Add developer section, 2022-12-13).

Ways to solve that:
- [ ] Disallow assertions in inopportune times, like data generators, hook into EvaluateResults maybe
- [ ] Treat each data generator as testcase, and group those into a pre-setup suite. Don't enforce assertions in data generators though. Don't start the test run at all if that pre-setup test suite has assertion errors. Needs to output proper JUNIT/TAP. These special test cases also ignore all test case hooks.
- [ ] Collect assertion failures in data generators and attribute them to the respective test cases. This is a 1:N mapping as multiple testcase can have the same generator.

Garados007:
Each data generator is now wrapped as a virtual test case which is also reported into the test results. To achieve this the setup for data generator has been changed into three phases:

1. Loading the list of data generate and do some general checks (for
   example if the data generator can be found or its function signature
   is correct).
2. Execute all data generators one by one and wrap each of them in their
   own test case. Do some basics checks with the generated results (for
   example if an invalid wave is returned).
3. Assign these data generators to their dependend test cases. Do some
   checks (for example if the types of the generated results match the
   function signature of the test case).

Close #353

-----

This PR does also consists of two refactoring steps. In one refactoring commit are all data generator, multi data and multi-multi data test related functions moved in their own file which reduces the size and complexity of UTF_Basics a lot. The second refactoring commit moves some polyfill methods into their own file.


Garados007:

MichaelHuth (changes requested):
I have not in depth reviewed the code, but there are generally some issues:

Is the behavior change for invalid data generators or their results intended?
Now the IUTF markes the related TC as skipped, before we aborted.

The original idea was that an invalid data generator indicates a wrong test setup, where the IUTF aborted with an error message.
Now, with the test case skipped this result is part of the data generator test suite and is only getting visible after everything ran through, which could take some hours worst case.

If these checks in phase 3 fail, we could just xml-close the junit file and don't run the aftual test cases. This would be closest to the previous abort.

Also if I have a data generator that returns a null wave, we get an internal IUTF runtime error in the first actual test case.
If you add to Various:
```igorpro
static Function/WAVE DgenNullWave()

	WAVE wv = $""
	
	return wv
End

// UTF_TD_GENERATOR DgenNullWave
static Function TestDgenNullWave([val])
	variable val
	
	PASS()
End
```
you get
```
  Data Generator function UTF_Various#DgenNullWave returns a null wave.
  Note: In test case UTF_Various#TC_MMD_ZeroSize data generator function (UTF_Various#GeneratorZeroSize) returns a zero sized wave. Test case marked SKIP.
  Note: In test case UTF_Various#TC_MD_ZeroSize data generator function (UTF_Various#GeneratorZeroSize) returns a zero sized wave. Test case marked SKIP.
  Data Generator function UTF_Various#DgenNullWave has no suitable data for test case UTF_Various#TestDgenNullWave. Test case marked SKIP.
  Start of test "Unnamed"
  Entering test suite "Various.ipf"
  Entering test case "TriggerWarningToHistory"
  Fatal: Internal runtime error in UTF 748:"WAVE;WAVE reference to "wv" failed." before executing test case "TriggerWarningToHistory".
  Leaving test case "TriggerWarningToHistory"
  Finished with no errors
  Leaving test suite "Various.ipf"
  Test finished with 2 errors
  End of test "Unnamed"

```

That happens for data generators that return a 2D wave as well, with an RTE in GetTestCaseCount, where it seems to be an incosistency between the data generator list in the test run setup and the actual present data generator waves in the wave reference wave.

Garados007:
@MichaelHuth Please re-review this.

t-b:
Assigning @MichaelHuth as I think this was the intention.

procedures/unit-testing-test-md-gen.ipf
```diff
@@ -245,3 +270,71 @@ static Function/WAVE CheckDGenOutput(procWin, fullFuncName, dgen, wType0, wType1
...
+			endif
+
+			FUNCREF TEST_CASE_PROTO_DGEN fDgen = $dgen
+			if(!UTF_FuncRefIsAssigned(FuncRefInfo(fDgen)))
+				fullFuncName = testRunData[i][%FULLFUNCNAME]
+				sprintf msg, "Data Generator function %s has wrong format. It is referenced by test case %s.", dgen, fullFuncName
+				UTF_Reporting#ReportErrorAndAbort(msg)
+			endif
+
+			succeed = 1
```
MichaelHuth, line 310: I think you don't need a success state variable here, that is also initialized to success.
The information you want to track is if you add the data generator or not to the wave.

procedures/unit-testing-test-md.ipf
```diff
@@ -97,10 +135,7 @@ static Function CheckFunctionSignatureTC(procWin, fullFuncName, dgenList, markSk
 	endif
 
 	dgen = UTF_Test_MD_Gen#GetDataGenFullFunctionName(procWin, fullFuncName)
-	WAVE wGenerator = UTF_Test_MD_Gen#CheckDGenOutput(procWin, fullFuncName, dgen, wType0, wType1, wRefSubType)
-
-	dgenList = AddListItem(dgen, dgenList, ";", Inf)
-	UTF_Test_MD_Gen#AddDataGeneratorWave(dgen, wGenerator)
+	WAVE/Z wGenerator = UTF_Test_MD_Gen#CheckDGenOutput(fullFuncName, dgen, wType0, wType1, wRefSubType)
```
MichaelHuth, line 138: /Z is not needed here

tests/TestResultsTests/GeneratorTests.ipf
```diff
@@ -0,0 +1,110 @@
...
+	variable var
+
+	string expect, result
+	variable childStart, childEnd, startTime, endTime
+
+	CHECK_EQUAL_VAR(1337, var)
+
+	WAVE/T/Z tc = Utils#LastTestCase(tcName = "TS_GeneratorTests#GeneratorEmpty", globalSearch = 1)
+	INFO("Bug: test case not found")
+	REQUIRE(WaveExists(tc))
```
MichaelHuth, line 29: We have REQUIRE_WAVE as assertion for such check

procedures/unit-testing-basics.ipf
```diff
@@ -733,6 +736,32 @@ static Function CreateTestRunSetup(procWinList, matchStr, enableRegExp, errMsg,
...
+		dgenList = testRunData[i][%DGENLIST]
+
+		if(UTF_Utils#IsEmpty(dgenList))
+			continue
+		endif
+
+		procWin = testRunData[i][%PROCWIN]
+		fullFuncName = testRunData[i][%FULLFUNCNAME]
+
+		if(UTF_Test_MD#CheckFunctionSignatureTC(procWin, fullFuncName, markSkip))
```
MichaelHuth, line 753: I think we should change the old behavior of skipping the TC in favor of aborting, though I am open to discuss that as I might miss an application for this behavior.

@t-b what do you think?

t-b: @MichaelHuth 
> I think we should change the old behavior of skipping the TC in favor of aborting, though I am open to discuss that as I might miss an application for this behavior.

Can you elaborate?

MichaelHuth: I was wondering if this behavior is used in some real world case.
Because with the skip behavior an accidental wrongly assigned data generator to an test case where the signature does not fit results in silently skipping that test case.
I would rate that as a wrong test setup, where the IUTF should abort.

t-b: > I would rate that as a wrong test setup, where the IUTF should abort.

Totally agree!

Garados007: I cannot remove the skip in `CheckFunctionSignatureTC` completely. `CheckDataGenZeroSize` is setting the test case as skip if the data generator returned an empty wave (which is not an error).

Garados007: All other conditions for setting test cases are now removed and changed into an abort.

Garados007:
@MichaelHuth Ready for review.

MichaelHuth (changes requested):
Just a minor comment, the rest of the current code looks good so far.

There is one thing missing:
When executing the data generators as test cases you have to set / reset the IgorDebugger as with the regular test cases. Otherwise Igor will e.g. stop with the Debugger on an RTE if DebugOnError is enabled.

t-b:
@MichaelHuth Just to be sure that this works now: Did you try to run one MIES hardware test with this PR?

MichaelHuth:
I did not wrote that it works now.
I was about to do some MIES tests as final test.

@Garados007 It turns out that the setup for test cases for the data generators is not complete as I get an IUTF internal RTE when using an assertion in a data generator function from a non-existing `igor_debug_assertion` in `DebugFailedAssertion`.

If the global is created the MIES HW tests start running...

t-b:
> I did not wrote that it works now.
> I was about to do some MIES tests as final test.

Thanks, sorry for being impatient.

Garados007:
I have refactored the debug helper methods into their own file. This allowed me to copy the set/reset of the Igor Debugger from regular test cases to the data generators. During this fixed also the bug with the non-existent `igor_debug_assertion` as it is always created before each data generator execution.

Garados007:
@MichaelHuth Ready for review.

procedures/unit-testing-test-md-gen.ipf
```diff
@@ -191,57 +220,132 @@ static Function EvaluateDgenTagResult(err, fullFuncName, varName)
...
+			procWin = StringByKey("PROCWIN", FunctionInfo(dgen))
+
+			FUNCREF TEST_CASE_PROTO_DGEN fDgen = $dgen
+			if(!UTF_FuncRefIsAssigned(FuncRefInfo(fDgen)))
+				fullFuncName = testRunData[i][%FULLFUNCNAME]
+				sprintf msg, "Data Generator function \"%s\" has wrong format. It is referenced by test case \"%s\".", dgen, fullFuncName
+				UTF_Reporting#ReportErrorAndAbort(msg)
+			endif
+
+			UTF_Debug#SetDebugger(debugMode)
```
MichaelHuth, line 308: Setting the Debugger at runtime is rather slow. For the regular TCs we do that is TestBegin/TestEnd.
Please don't set the Debugger for each DGEN-TC, it is sufficient to set/restore it outside the for loops here.

MichaelHuth (changes requested):
The MIES hardwaretests work with this changes.

The point where the debugger is set/reset is a performance blocker, see comment.

procedures/unit-testing-test-md-gen.ipf
Garados007: Thanks, I fixed it.

MichaelHuth (approved):
ok

t-b:
Thanks both!

PR #392, opened 2023-01-24
Fix user hook issues #384 #385

### LINKED ISSUE #384: TEST_CASE_END user hook but no TEST_CASE_BEGIN user hook for skipped test cases
```igorpro
#include "unit-testing"

Function TEST_CASE_BEGIN_OVERRIDE(string name)
	printf "hook begin: %s\n", name
End

Function TEST_CASE_END_OVERRIDE(string name)
	printf "hook end: %s\n", name
End

// UTF_SKIP
Function TestCase()
	print "test case"
	PASS()
End
```
```
•RunTest("Procedure")
  Start of test "Unnamed"
  hook end: TestCase
  Finished with no errors
  Test finished with no errors
  End of test "Unnamed"
```

### LINKED ISSUE #385: AfterTestCaseUserHook never executed if user hook failed with RTE or Abort
Is this intended behavior? AfterTestCaseUserHook is responsible for the following things:

- cleanup work data folder
- reset data folder to last data folder
- wave tracking leak checks

Garados007:
1. Execute AfterTestCaseUserHook independent from user hook

    The execution of AfterTestCaseUserHook was dependent of the result of
    the user hook. If the user hook failed for some reasons the
    AfterTestCaseUserHook was never executed. This has been fixed that
    AfterTestCaseUserHook is always executed independent of the user hook
    result.
2. Remove unused ExecuteUserHook return value

    After 38133d6 (Execute AfterTestCaseUserHook independent from user hook,
    2023-01-24) the return value of ExecuteUserHook is no longer used in the
    code. As such that it is removed completely from this function.
3. Skip test case end user hook if test case skipped

    The user hooks for test case begin and end are now only then executed if
    the test case itself is planed for execution.

Fix #384
Fix #385

### DIFF
--- a/procedures/unit-testing-hooks.ipf
+++ b/procedures/unit-testing-hooks.ipf
@@ -178,9 +178,8 @@ static Function ExecuteHooks(hookType, hooks, enableTAP, enableJU, name, procWin
 			AfterTestCase(name, skip)
 			FUNCREF USER_HOOK_PROTO userHook = $hooks.testCaseEnd
 
-			if(!ExecuteUserHook(name, userHook, procWin, HOOK_LEVEL_TEST_CASE))
-				AfterTestCaseUserHook(name, param)
-			endif
+			ExecuteUserHook(name, userHook, procWin, HOOK_LEVEL_TEST_CASE)
+			AfterTestCaseUserHook(name, param)
 
 			if(!skip)
 				// finalize the normal test case at tcOutIndex and reset the test case index to the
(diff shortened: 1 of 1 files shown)


t-b:
Review:

In 		`case IUTF_TEST_CASE_BEGIN_CONST:` we still have

```
			if(!skip)
				TestCaseBegin(name)
				ExecuteUserHook(name, userHook, procWin, HOOK_LEVEL_TEST_CASE)
			endif
```

which I think is also something we want to fix as `TestCaseBegin` should also always be executed.

Garados007:
@t-b 
> which I think is also something we want to fix as TestCaseBegin should also always be executed.

This could be part of a new issue. TestCaseBegin is executing the following things:

- Create new unique temp datafolder
- log to output that a test case is started

I don't think that we need this for skipped test cases (especially the log output).

t-b:
@Garados007 Thanks, makes sense.

t-b (approved):

PR #393, opened 2023-01-24
Require optional function parameter for reentry

### LINKED ISSUE #270: Tighten reentry function check in RegisterUTFMonitor
As Igor6 does not distinguish between optional and normal arguments, we should also add check for the reentry function for that. Similar to https://github.com/byte-physics/igor-unit-testing-framework/pull/268

Garados007:
The function parameter are now checked if they are all optional as in the test cases. This fixes the loose function signature check in Igor 6 like it was done in 085e3d3 (Require optional parameter for test methods, 2022-10-07).

Close #270


Garados007:

t-b (approved):

PR #394, opened 2023-01-30
Use cached home path for accessing files

### LINKED ISSUE #287: Should not rely on fixed home path
When tests rename the current experiment and save it to a new folder we currently fail things like instrumentation. This is because the home path is currently assumed to be same at the start and end of test run.

We should create our own path to store home and use that instead.

Garados007:
At the start of RunTest the current home directory of the experiment file is fetched and cached in a global string. If for some reasons a test case changed to current home directory (for example storing the experiment file elsewhere) all UTF functions will still use the old home directory.

Close #287

### DIFF
--- a/procedures/unit-testing-autorun.ipf
+++ b/procedures/unit-testing-autorun.ipf
@@ -18,13 +18,17 @@ End
 ///
 /// @returns one of @ref AutorunModes
 Function GetAutorunMode()
-	GetFileFolderInfo/Q/Z/P=home "DO_AUTORUN.TXT"
+	string path
+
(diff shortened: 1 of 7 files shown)


Garados007:

MichaelHuth (changes requested):
Yes, but how about wrapping the whole application of the homepath in a function like:
AtHome(string fileName, [variable unusedName])

Garados007:
@MichaelHuth I have implemented your suggestion. The commit has changed a bit. Please review again.

MichaelHuth (approved):
Looks good to me.

PR #399, opened 2023-02-03
.gitattributes: Ignore the documentation folder for the language stat…

t-b:
…istics

See https://github.com/github/linguist/blob/master/docs/overrides.md#documentation.


Garados007:
Didn't know there was a such a thing, but okay, I approve.

PR #407, opened 2023-02-18
Fix: Replace if(strlen(str)) with UTF_Utils#IsEmpty

### LINKED ISSUE #303: Replace if(strlen()) with if(UTF_Utils#IsEmpty()) if suitable
In some places the statement `if(strlen(str))` is used to check if a string is empty or not. This has a design flaw that it will throw an RTE if the given string is null. It is better to use `if(UTF_Utils#IsEmpty())` because it will also check these cases. The goal is not to replace all `strlen()` only these where it is necessary.

Garados007:
Strings from function arguments can be null and should be untrusted.
Using if(strlen(str)) to the check if its empty is therefore insecure
and prone to errors. These calls have been replaced with the better
UTF_Utils#IsEmpty.

Not all if(strlen(str)) are insecure and therefore only the mentioned
case is updated. If the string originate from a wave or an SVAR, which
cannot be null, the code if(strlen(str)) is kept and not replaced.

Close #303

### DIFF
--- a/procedures/igortest-reporting.ipf
+++ b/procedures/igortest-reporting.ipf
@@ -205,7 +205,7 @@ static Function AddError(message, type, [incrErrorCounter])
 		wvTestCase[%CURRENT][%STATUS] = type
 		wvTestCase[%CURRENT][%NUM_ASSERT_ERROR] = num2istr(str2num(wvTestCase[%CURRENT][%NUM_ASSERT_ERROR]) + 1)
 	endif
-	if(strlen(message))
+	if(!IUTF_Utils#IsEmpty(message))
 		wvTestCase[%CURRENT][%STDERR] = AddListItem(message, wvTestCase[%CURRENT][%STDERR], "\n", Inf)
 	endif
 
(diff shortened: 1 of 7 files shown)


Garados007:

t-b:
> Not all if(strlen(str)) are insecure and therefore only the mentioned
case is updated. If the string originate from a wave or an SVAR, which
cannot be null, the code if(strlen(str)) is kept and not replaced.

I would actually prefer if all strlen() on plain strings which are used for checking empty/filled are replaced with `IsEmpty`. We could even move to `IsEmpty` taking a copy so that this can be used with structure elements and SVARs as well. We have done that in MIES already, see AllenInstitute/MIES@9857c4bf (Changed IsEmpty to not use string by reference, 2021-09-14)

Garados007:
Okay, I will update `IsEmpty` and it's references.

procedures/igortest-reporting.ipf
```diff
@@ -513,10 +513,12 @@ threadsafe static Function IUTF_PrintStatusMessage(msg, [allowEmptyLine])
 	variable allowEmptyLine
 
 	string tmpStr
+	variable len
 
 	allowEmptyLine = ParamIsDefault(allowEmptyLine) ? 0 : !!allowEmptyLine
 
-	if(!allowEmptyLine && strlen(msg) == 0)
+	len = strlen(msg)
+	if(!allowEmptyLine && (numtype(len) == 2 || len <= 0))
```
MichaelHuth, line 521: Why not use IsNullOrEmpty here?

Garados007:
@MichaelHuth 
> The behavior changes need some documentation.

What kind of documentation do you think of? The function documentation has been updated, the commit message mention it and `IsEmpty` is nowhere used or mentioned in our sphinx documentation. Should I add something to the changelog?

`IsEmpty` was an internal function all the time and is hidden with `///@cond HIDDEN_SYMBOL`. I don't think there are external user.

procedures/igortest-reporting.ipf
Garados007: I have updated the documentation as discussed yesterday.

Garados007:
Removed `IsNullOrEmpty` as discussed yesterday.

t-b (approved):
LGTM.

MichaelHuth (approved):
Looks good.

PR #409, opened 2023-02-21
Fix long wave labels

### LINKED ISSUE #390: Bug: Long full function names as wave labels
We use at different locations in our code full function names as wave labels to identify their entries faster. This yield to some problems if the full function name is very large and larger than the allowed limit for the current IP version. This happened in #378 and was fixed with 0309a67 (Function Tags: Fix long function name bug, 2023-01-16) for function tags.

At least I found the following waves that use full function names as dimension label:

- `root:Packages:UnitTesting:DataGeneratorWaves` is a wave-reference-wave for the results of the data generators
- `root:Packages:UnitTesting:MMDataState` is a two dimensional value wave for storing test-states of multi-multi-data test cases

Garados007:
MMDataState and DataGeneratorWaves used the full function names as wave label to find its
entries. This can lead to bugs when the full function name becomes
larger than the supported wave label size. This is fixed using a second
text wave to store the function names at the same index as the entries
in MMDataState or DataGeneratorWaves.

This is the same fix as in 0309a67 (Function Tags: Fix long function
name bug, 2023-01-18).

Close #390

### DIFF
--- a/procedures/igortest-basics.ipf
+++ b/procedures/igortest-basics.ipf
@@ -638,7 +638,7 @@ End
 static Function GetTestCaseCount([procWin])
 	string procWin
 
-	variable i, j, size, dgenSize
+	variable i, j, size, dgenSize, index
 	variable tcCount, dgenCount
 	string dgenList, dgen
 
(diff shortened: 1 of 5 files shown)


Garados007:

PR #410, opened 2023-02-21
Doc: behavior of RTE/Abort in DG and UH

### LINKED ISSUE #387: Document behavior of aborts/RTE inside user hooks
Suitable location: https://docs.byte-physics.de/igor-unit-testing-framework/advanced.html#test-hooks

Garados007:
3bdcaaf (Introduce data generators as virtual test cases, 2023-01-26)
and 0a18141 (Treat user hooks as test cases for the results, 2023-01-13)
changed the behavior of pending RTE and aborts at the end of data
generator and user hooks. The documentation has been updated to reflect
this behavior.

Close #387

### DIFF
--- a/docu/sphinx/source/advanced.rst
+++ b/docu/sphinx/source/advanced.rst
@@ -79,7 +79,10 @@ cleanup is done afterward. The *next* Test Case then starts with the data the
    By default the Igor debugger is disabled during the execution of a test run.
 
 Assertions can be used in test hooks. However it is enforced by the IUTF that
-the test case itself must contain at least one assertion.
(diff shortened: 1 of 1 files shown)


t-b:
Review:

English: `as this is considered as an invalid test setup.` -> `as this is considered an invalid test setup.`

> If an assertion in a test hook fails the test run is still executed normally.

This describes the status quo with CHECK/WARN but I would think this is not true for REQUIRE or?

Garados007:
> This describes the status quo with CHECK/WARN but I would think this is not true for REQUIRE or?

Yes. I will try to clarify this.

t-b (approved):

PR #413, opened 2023-02-22
Allow rerun of flaky test cases

### LINKED ISSUE #164: RunTest should allow to rerun everything X times

Garados007:
Some test cases can fail from time to time and only need to be rerun
some times to succeed.

Close #164


Garados007:

t-b:
Review:

Nice addition, works mostly!

- [ ] Can we have some more tests with multi data and mulit multi data?

I've tried with

```
Function/WAVE GetWaveA()

	Make/FREE wv = {100, 200}
	
	return wv
End

Function/WAVE GetWaveB()

	Make/FREE/T wv = {"a", "b"}

	return wv
End

// IUTF_RETRY_FAILED
// IUTF_TD_GENERATOR v0:GetWaveA
// IUTF_TD_GENERATOR s0:GetWaveB
static Function UnreliableStatus([STRUCT IUTF_mData &m])
	NVAR counter = root:retryCounter
	counter += 1
	CHECK_GE_VAR(counter, 5)

	print m.v0, m.s0
End
```

which works.

- [ ] Can we document the behaviour for multi and multi-multi test cases? I think the current behaviour of not incrementing the dgenIndex on retry is good, but that needs to be written down. @MichaelHuth Please chim in if you object.
- [ ] Can we make it work with REQUIRE as well?
- [ ] `CHECK(counter >= 5)` -> `CHECK_GE_VAR(counter, 5)`
- [ ] Can you check in the tests that test case begin/end hooks are executed once per retry?

Garados007:
Applied the requested changes.

> Can we make it work with REQUIRE as well?

Yes. I have added the support for this. It was a bit tricky as failed REQUIREs works differently from a failed CHECK as it aborts the execution right away. I made the support for rerunning failed REQUIREs optionally as there are other conditions that could abort the test case. I updated the documentation and recommend the use of CHECKs for flaky conditions.

t-b (approved):

PR #417, opened 2023-02-24
 Fix order of test case execution + Add shuffling of execution order

### LINKED ISSUE #375: Allow to randomize the test order in a test suite
- [x] Needs to use the best PRNG available for each IP version
- [x] RunTest: optional parameters, defaults to off
- [ ] Might be worth looking at InPlaceRandomShuffle in MIES

### LINKED ISSUE #416: Order of test case execution
I found something interesting when I took some preparations for #375 and I am not sure if it's intended or not. Let's say we have two procedure files `Proc0` and `Proc1` with mostly the same code:

```IgorPro
#pragma rtGlobals=3				// Use modern global access method and strict wave access
#pragma DefaultTab={3,20,4}		// Set default tab width in Igor Pro 9 and later
#pragma ModuleName=Proc0 // procedure Proc1 has the module name Proc1

#include "igortest"

static Function TestA()
	PASS()
End

static Function TestB()
	PASS()
End
```

If I execute `RunTest("Proc0;Proc1", testcase="TestA;TestB")` I get the following `TestRunData`:

grafik

I would have expected that IUTF would execute all test cases from `Proc0` first and then `Proc1` but it seems like the test cases are executed out of order.

I don't know it this is intended behavior so I am asking for it.

Garados007:
The order of test case execution is now fixed and IUTF now always
executes the test cases from the same test suite together. The execution
of the test suites are now in that order they are listed in the RunTest
call (or alphabetically if regex used).

The code of CreateTestRunSetup is now much simpler and contains less
nested loops.

Close #416

---

The user can now shuffle the execution order of test suites and test
cases. It is also possible to opt-out shuffling test cases in specific
test suites.

The default execution order is documented alongside the new feature.

Close #375


Garados007:

t-b:
Nice one!

The shuffling options look good, I also like both levels (test suite and cases). The idea to skip a test suite is also a good.

Comments:
- Unit test for CreateTestRunSetup, for each commit, would be nice
- Can we limit the search range for the procedure file tag to the top ~20 lines? The main idea would be to encourage good code and also make it faster.
- IUTF_BAN_SHUFFLE_TEST_CASE -> IUTF_NO_SHUFFLE_TEST_CASES ?
- Does your new `HasProcedureShuffleBan` function work in an IM? GetFunctionTagWave does some magic with ProcedureText to achieve that
- I don't like that we now have an additional place for the tag regexp. So either, at your choice, create a function which returns the tag regexp given the tag or extend GetFunctionTagWave to also fetch procedure tags. Your new regexp also leaves the `\\Q`/`\\E` out.

t-b (changes requested):
See comment

Garados007:
Fixed.

And who could have guessed? Testing randomness in Igor is pretty ... unpredictable. 😂 I needed special test cases for Igor 7 and 8. 🤷

t-b:
Very nice solution with `IsTagMatch`.

Some comments regarding the tests:

```
+    // Restore seed to a "random" one
+    seed = DateTime * 1e-6
+    SetRandomSeed seed - floor(seed)
```

You don't need that. The random seed after a test case is not something anyone should care about.

TestHelper:

That helper looks mostly good. But it misses to check two crucial things. It currently passes when passing an empty `expect` wave. And it also passes if TestRunData has more entries than `expect`.

So I would add

```
CHECK_EQUAL_VAR(length, DimSize(wv, UTF_ROW))
CHECK_GT_VAR(lenth, 0)
```

to nail that down.

Always when writing a loop in a test, you need to ask yourself, "Is it okay if the loop is not executed at all?". And in many cases the answer is no, and thus the above checks are required.

Garados007:
Fixed.

t-b (approved):

PR #418, opened 2023-02-24
Add *_RTE, *_ANY_RTE and *_NO_RTE assertions

### LINKED ISSUE #339: Add assertion for checking pending RTEs
Introducing an assertion like this can make many RTE checks easier and in less code:

```igorpro
// current approach
Function ShouldRTE()
	try
		ProblematicFunction(invalidInput); AbortOnRTE
		// RTE was not thrown - this is unexpected
		INFO("RTE was not thrown")
		CHECK(0)
	catch
		string message = GetRTErrMessage()
		variable err = GetRTError(1)
		INFO("Invalid error (%d): %s", n1=err, s1=message)
		CHECK_EQUAL_VAR(1, err) // 1 is the expected error code
	endtry

	// continue ...
End
```

```igorpro
// with the new CHECK_RTE method
Function ShouldRTE()
	INFO("ProblematicFunction is expected to fail with 1 when using invalidInput")
	ProblematicFunction(invalidInput)
	CHECK_RTE(1) // 1 is the expected error code

	// continue ...
End
```

`CHECK_RTE` is always checking if the error code matches the provided one. If it doesn't it produce an assertion error.

`CHECK_ANY_RTE` is checking for any error code. It produces only an assertion error if no error is pending.

`CHECK_NO_RTE` is checking that no pending error exists. It produces only an assertion error if a pending error exists.

All functions are clearing the RTE at the end.

If one check failed it produces an assertion error. The resulting assertion message should contain the RT error code and the RTE message.

We need new reference code examples in the user documentation, see also https://docs.byte-physics.de/igor-unit-testing-framework/examples.html?highlight=pass#example8.

Garados007:
These assertions check the current RTE state and create an error if its
unexpected. This simplifies the user code to check for RTEs during
runtime.

Close #339

### DIFF
--- a/docu/examples/example8-uncaught-runtime-errors.ipf
+++ b/docu/examples/example8-uncaught-runtime-errors.ipf
@@ -7,12 +7,12 @@
 
 Function TestWaveOp()
 
-		Wave wv = $""
-		print wv[0]
-		print "This will be printed, even if a RTE occurs."
+	Wave wv = $""
+	print wv[0]
+	print "This will be printed, even if a RTE occurs."
 
-		WAVE/Z/SDFR=$"I dont exist" wv; AbortOnRTE
-		print "This will not be printed, as AbortOnRTE aborts the test case."
+	WAVE/Z/SDFR=$"I dont exist" wv; AbortOnRTE
+	print "This will not be printed, as AbortOnRTE aborts the test case."
 End
 
 Function TestWaveOpSelfCatch()
--- a/docu/examples/example8-uncaught-runtime-errors.ipf
+++ b/docu/examples/example8-uncaught-runtime-errors.ipf
@@ -15,17 +15,30 @@ Function TestWaveOp()
 	print "This will not be printed, as AbortOnRTE aborts the test case."
 End
 
-Function TestWaveOpSelfCatch()
+Function CheckForRTEs()
 
+	Wave wv = $""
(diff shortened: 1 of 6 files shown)


Garados007:

t-b:
Very nice addition!

The assertions work as expected. I do have some questions about the example though ;)

- TestWaveOp: Whitespace in separate commit would be nice. But why remove `AbortOnRTE`? And if that is the right thing to do the print message should not mention it anymore.
- TestWaveOpSafe: This seems to be a better way to do things. Can we maybe rename that test case? And I'm wondering why we still need to have the try/catch? With the new assertions I would just do

```igorpro
	INFO("checks if CustomUserFunction returns with no RTE")
	CustomUserFunction()
	CHECK_NO_RTE()
```
or?

t-b (changes requested):
See comment

Garados007:
> Whitespace in separate commit would be nice. 

👍 

> But why remove AbortOnRTE?

I forgot to put it back after experimenting.

> And I'm wondering why we still need to have the try/catch?

At this part I want to show that the user can now handle RTE and aborts differently. Previously, the user has to add `AbortOnRTE` and handle both in the `catch`-block.

Garados007:
Fixed it.

t-b (approved):

PR #428, opened 2023-03-06
Add keyword elseif for complexity analysis

Garados007:
c36581a (Add complexity calculation for cobertura output, 2023-03-06)
didn't add this.


Garados007:
I found this bug while browsing your MIES tests. This fix can increase the complexity numbers a bit. 😅

PR #433, opened 2023-03-13
Clear abort flag at the begin of RunTest

### LINKED ISSUE #431: F2 vs F4 aka abortflag is not intialized

Garados007:
This prevents that an abortflag from a previous RunTest has some
influence for data generator of the current RunTest. Previously, the
abort flag is cleared at the test run begin hook which is called after
the data generator.

Close #431


Garados007:
@t-b @MichaelHuth The pipeline is not finished yet but I think it's ready for review.

t-b (approved):

PR #443, opened 2023-03-22
GenerateDimLabelDifference: Don't ignore case

t-b:
We use EqualWaves for comparison of dimension labels and that does it since forever case sensitive.

So we have to also output a difference if the labels only differ in case. Bug introduced in the initial version in 493779da (EQUAL_WAVE_WRAPPER: Work around EqualWaves dimension labels bug, 2019-03-12).

- [x] Add test, VTE?


t-b:
Please have a look.

MichaelHuth (approved):
Looks good.

Though, the CI pipeline seems to have some issue. (probably unrelated)

MichaelHuth (approved):
LGTM

PR #470, opened 2024-06-06
Add SKIP_TESTCASE and SKIP_TESTCASE_EXPECT_FAILS assertions

### LINKED ISSUE #467: Add function to skip a test case
Add function like `IUTF_SKIP()` that aborts the test case immediately and treats it as skipped. The function should output INFO(...) text if available.

Use case: If a test case detects that e.g. hardware requirements are not met in a specific setup configuration then it can be setup to be skipped.

MichaelHuth:
Both assertion abort the current testcase. Any registered reentry function is unregistered.

When SKIP_TESTCASE_EXPECT_FAILS is called, previously failed assertions in the same testcase are treated as expected failures.

close #467

### DIFF
--- a/procedures/igortest-assertions.ipf
+++ b/procedures/igortest-assertions.ipf
@@ -30,6 +30,40 @@ Function FAIL()
 	IUTF_Wrapper#TRUE_WRAPPER(0, REQUIRE_MODE)
 End
 
+/// Skips current test case
+///
+/// The test case is aborted.
+/// Already failed assertions are treated as expected failure.
+/// Any registered reentry is automatically unregistered.
+Function SKIP_TESTCASE_EXPECT_FAILS()
+	IUTF_Basics#SetExpectedFailure(1)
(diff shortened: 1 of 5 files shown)


Garados007 (changes requested):
I am missing some test cases that call these new assertions and check if

1. the output in the log is as expected (you are changing the order of the info output calls, so I would expect a test for this).
2. the test results are correctly stored in the global waves. Doing this ensures that igortest handles the new assertions correctly and the output (e.g. junit) is as expected.

MichaelHuth:
I did not change the order of the output, the output is just moved to an own function.

I added some tests that check if the TCs are correctly counted as skipped.

Garados007:
I am just curious. Wouldn't `tests/TestResultsTests/EasyTests.ipf` be a more fitting place for the tests than `tests/Various.ipf`?

MichaelHuth:
Ok I moved it.

Garados007 (approved):

Garados007:
LGTM

@t-b It seems like you have to approve this PR as code owner

PR #476, opened 2024-07-22
Performance Increase

### LINKED ISSUE #472: Clearing of info wave is slow
https://github.com/byte-physics/igortest/blame/d557db885f2077bb1998cf9ef938477881c44775/procedures/igortest-reporting.ipf#L271

Some test profiling for UTF_Labnotebook.ipf from MIES shows that this line uses 15% of all function time for the whole test suite.

@Garados007 Is the clear really necessary or would setting the vector length suffice ?

### LINKED ISSUE #474: Perf: EQUAL_WAVE_WRAPPER generates the msg string for each mode and evaluates the result afterwards
The check for CHECK_EQUAL_WAVES is slow because we generate the detailed message before we evaluate the result.

Look into that, if the performance can be increased by not creating output strings that are not output in the end.

MichaelHuth:
These changes increase the performance of the IUTF.
For 1000x CHECK_EQUAL_WAVES for the success case the rough numbers are:

Before: 7.37 s
After: 0.31 s

- [x] remove debugoutput functionality

close #474 
close #472

### DIFF
--- a/docu/sphinx/source/code.rst
+++ b/docu/sphinx/source/code.rst
@@ -9,5 +9,4 @@ converted for `sphinx <http://www.sphinx-doc.org>`__ using `breathe
    :maxdepth: 2
 
    assertions
-   helpers
    flags
--- a/docu/sphinx/source/helpers.rst
+++ b/docu/sphinx/source/helpers.rst
@@ -1,8 +1,0 @@
-Helper Functions
-----------------
-
-
-The following Helper Functions are available:
-
-.. doxygengroup:: Helpers
-   :content-only:
--- a/procedures/igortest-assertion-wrappers.ipf
+++ b/procedures/igortest-assertion-wrappers.ipf
@@ -735,7 +735,8 @@ static Function EQUAL_WAVE_WRAPPER(wv1, wv2, flags, [mode, tol])
 	variable mode, tol
 
 	variable i, result
-	string str, detailedMsg, name1, name2
+	string detailedMsg, name1, name2
+	string str = ""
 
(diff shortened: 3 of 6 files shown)


Garados007:
The numbers are astounding. 😱

The changes looks good and can be merged.

Garados007 (dismissed):

Garados007:
@MichaelHuth Can you link the respective issues, so they will be automatically closed?

MichaelHuth:
@Garados007 I was also wondering about the DebugOutput function that is called for each result evaluation. This calls EnabledDebug that does a GetPackageFolder resolve followed by an NVAR resolve. Both is rather slow.

Do you think it would be ok for the debug output functionality to wrap it through a #define instead of a global variable ?

Garados007:
I checked the implementation in igortest. This functionality is never used and meant to be enabled by the user using `EnableDebugOutput()`. The gain is also quite limited, because this will only print some data to the history that is already available in the result waves (and the result waves contain more data). This might be an artifact of the time when it was implemented and I have never made use of it.

I assume it is okay if you exchange the global variable with a `#define`. And I am also okay if you remove this functionality entirely, although I think we have to discuss this with Thomas first.

t-b:
Decision from todays meeting: Throw out debug mode, also from documentation.

MichaelHuth:
It seems it was nowhere documented. I could not find a mention in the rst files.

Garados007 (approved):

Garados007:
@t-b I cant enforce the merge because of failing CI. Can you merge this?

PR #483, opened 2024-09-18
Fix sphinx documentation style

Garados007:
We have updated our used sphinx and alabaster versions and they had some major breaking changes in the style and our current documentation looks a bit off. These changes ensures that we have the same style like before.


Garados007:
I can't merge this myself.

MichaelHuth (approved):

PR #493, opened 2025-04-22
Tracing: Fix switch statement tracing to support IP10

### LINKED ISSUE #492: Tracing is broken with IP10
The following function 

```igorpro
Function  func()

	variable val

	switch(val)
		// some comment
		case 1:
		break
		// other comment
		case 2:
		break
#ifdef AUTOMATED_TESTING
		case 3:
		break
#endif		
	endswitch
End
```

can not be traced anymore as we add `Z_` commands in places which trigger

```
This code will never execute because it is before a case or default statement or is after a break statement inside a switch or strswitch group.
```

Code with tracing:

```igorpro
Function  func()
Z_(0, 6)
Z_(0, 22)

Z_(0, 8)
	variable val

Z_(0, 10)
	switch(val)
Z_(0, 11)
		// some comment
		case 1:
Z_(0, 12)
Z_(0, 13)
		break
Z_(0, 14)
		// other comment
		case 2:
Z_(0, 15)
Z_(0, 16)
		break
Z_(0, 17)
#ifdef AUTOMATED_TESTING
		case 3:
Z_(0, 18)
Z_(0, 19)
		break
#endif		
Z_(0, 20)
	endswitch
Z_(0, 21)
End
Function IUTF_TagFunc_903d205bbbbbde29157219a98e1ad6d48c86fadd7bcd26c461476d9122246469_IGNORE()
End
```

We need a not too hackish solution. The longterm plan would be to gain that capability in IPT.

Garados007:
With Igor 10 there is no code in specific regions of a switch statements allowed:

```igorpro
switch(condition)
    print "no code supported here"
    case VAR:
        print "normal code"
        break
        print "no code supported here"
endswitch
```

Our tracing logic added Z_ function calls in such regions and therefore is a
logic added that prevents the addition of such. This fix does the following
changes to the coverage report:

- All comments, that are located in regions that do not support the execution of
  Z_ functions (like certain regions in switch statements), are uncoverable.
- All compiler pragmas like #if and #else are now uncoverable.

Uncoverable lines are excluded from coverage report and will be rendered gray in
most visualizations.

This fix might change the overall code coverage slightly. This isn't the case if
you already had a code coverage of 100%.

Close #492

### DIFF
--- a/procedures/igortest-basics.ipf
+++ b/procedures/igortest-basics.ipf
@@ -4,6 +4,17 @@
 #pragma TextEncoding="UTF-8"
 #pragma ModuleName=IUTF_Basics
 
+#undef UTF_ALLOW_TRACING
+#if Exists("TUFXOP_Version")
+
+#if IgorVersion() >= 10.00
+#define UTF_ALLOW_TRACING
+#elif (IgorVersion() >= 9.00) && (NumberByKey("BUILD", IgorInfo(0)) >= 38812)
+#define UTF_ALLOW_TRACING
+#endif
+
+#endif
+
 ///@cond HIDDEN_SYMBOL
 
 static Constant FFNAME_OK        = 0x00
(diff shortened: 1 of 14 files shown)


Garados007:
Our tracing test succeeded with IP 10 beta 1 locally.

MichaelHuth (approved):
LGTM

PR #500, opened 2025-07-23
EvaluteRTE: Correctly report AbortOnValue aborts with negative abort codes

t-b:
We need to assume that all error codes not being -1, -2, -3, -4 are from AbortOnValue as we can't differentiate them. This now correctly reports AbortOnValue aborts with negative abort codes.

Broken since a7ab54c (More detailed RTE error message, 2017-03-07). It did not help that the Igor Pro documentation is also wrong in this regard.

### DIFF
--- a/procedures/igortest-basics.ipf
+++ b/procedures/igortest-basics.ipf
@@ -453,15 +453,14 @@ static Function EvaluateRTE(err, errmessage, abortCode, funcName, funcType, proc
 				IUTF_Reporting#AddError(str, IUTF_STATUS_ERROR)
 				break
 			default:
+				if(abortCode != 0)
+					sprintf str, "Encountered \"AbortOnValue\" Code %d in %s \"%s\" (%s)", abortCode, funcTypeString, funcName, procWin
+					IUTF_Reporting#AddFailedSummaryInfo(str)
+					IUTF_Reporting#AddError(str, IUTF_STATUS_ERROR)
+				endif
 				break
 		endswitch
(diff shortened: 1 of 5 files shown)


procedures/igortest-basics.ipf
```diff
@@ -453,15 +453,14 @@ static Function EvaluateRTE(err, errmessage, abortCode, funcName, funcType, proc
 				IUTF_Reporting#AddError(str, IUTF_STATUS_ERROR)
 				break
 			default:
+				if(abortCode != 0)
```
Garados007, line 456: This can be optimized. Just add a case which checks for 0 and does nothing and move the content of the if statement into the default case directly.

Doing this you have one level of indentation less and no if statement.

Garados007 (changes requested):

procedures/igortest-basics.ipf
t-b: Good point. But thinking about this again, the dheck for non-zero abort code is not correct either. The abort code can be anything. Fill send a fixed version.

MichaelHuth: Ok, so the -1, -2, -3 cases can possibly also be AbortOnValue then.

t-b: Unfortunately yes. My idea here is that this is not as likely as the other abort types.

t-b: Should be fixed.

